PR #21, opened 2017-04-27
Initial design contacts

clenrock:
Basics of the app. There is a main screen with a list of dummy names and a button to make a Bluetooth connection.

PR #20, opened 2017-04-26
Quality tools

TheBonheurs:
These changes should bolster our programming quality by adding support within gradle for plugins like checkstyle, findbugs, pmd and lint. All have their own customizable rulesets and print their reports in the report directory (within app/build folders).

PR #22, opened 2017-04-27
Implement Cucumber

jverbraeken:
This PR implements the Cucumber framework to verify if user stories are implemented.

Why? We want to be sure that all requirements of the client are implemented

How? Cucumber tests can be run by clicking "Edit Configurations" -> if "Android Tests" is not already added, add it by clicking the + -> Give the new Android Test Configuration is nice name, the module is app, the specific instrumentation runner is "com.nervousfish.nervousfish.test.Instrumentation" (without the quotes) -> click OK. Select the new configuration and click, after connecting your (virtual) Android device, on the Run button.

Feedback? The PR has an example of an Activity that has several user stories that are tested successfully. These examples do NOT have to be reviewed. These are only to simplify the development of Cucumber tests in the first week(s) and will be deleted when we have written our own Cucumber tests successfully. However, feedback about the existing files, made by @clenrock is appreciated.


.idea/vcs.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
TheBonheurs: According to @ericcornelissen this file is in the .gitignore so you should delete this manually

app/build.gradle
```diff
@@ -9,22 +9,48 @@ android {
         targetSdkVersion 25
         versionCode 1
         versionName "1.0"
-        testInstrumentationRunner "android.support.test.runner.AndroidJUnitRunner"
+        multiDexEnabled true
+
+        testApplicationId "com.nervousfish.nervousfish.test"
+        testInstrumentationRunner "com.nervousfish.nervousfish.test.Instrumentation"
     }
+
```
TheBonheurs: After #20 this section will change so you might need to manually merge stuff :)

app/src/androidTest/java/com/nervousfish/nervousfish/test/LoginActivitySteps.java
```diff
@@ -0,0 +1,104 @@
...
+    public void I_should_see_error_on_the_editTextView(final String viewName) {
+        int viewId = (viewName.equals("email")) ? R.id.email : R.id.password;
+        int messageId = (viewName.equals("email")) ? R.string.msg_email_error : R.string.msg_password_error;
+
+        onView(withId(viewId)).check(matches(hasErrorText(getActivity().getString(messageId))));
+    }
+
+    @Then("^I should (true|false) auth error$")
+    public void I_should_see_auth_error(boolean shouldSeeError) {
+        if (shouldSeeError) {
```
TheBonheurs: In lines 58, 59 you use a ternary operator ( cond ? resolve true : false) but continue here with a simple if () else(). I don't mind but maybe it is better for consistency sake to stick to one 😄

.idea/misc.xml
```diff
@@ -0,0 +1,46 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
ericcornelissen: Similar to `.idea/vcs.xml`, this file should not be in the PR 😄

app/build.gradle
```diff
@@ -9,22 +9,48 @@ android {
...
+        exclude 'LICENSE.txt'
+    }
+
+    lintOptions {
+        abortOnError false
+    }
 }
 
 dependencies {
-    compile fileTree(dir: 'libs', include: ['*.jar'])
```
ericcornelissen: Why did you remove this?

app/src/main/AndroidManifest.xml
```diff
@@ -19,6 +19,7 @@
                 <category android:name="android.intent.category.LAUNCHER" />
             </intent-filter>
         </activity>
+        <activity android:name=".LoginActivity"></activity>
```
ericcornelissen: Is this for the example tests? If so, add a comment for our future selfs so we know its not important

app/src/main/java/com/nervousfish/nervousfish/MainActivity.java
```diff
@@ -6,8 +6,6 @@
 import android.support.v7.app.AppCompatActivity;
```
ericcornelissen: The changes to the app will cause some issues with #21, I would say that one has priority over this one for now. After #21 is merged, you could add the login frame

app/build.gradle
```diff
@@ -9,22 +9,48 @@ android {
...
             proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
         }
     }
+
+    packagingOptions {
+        exclude 'LICENSE.txt'
+    }
+
+    lintOptions {
+        abortOnError false
```
ericcornelissen: Why did you add this, you didn't do anything related to linting (right)? Still, I would say we want to abort on a lint error (at least on Travis)

app/src/androidTest/java/com/nervousfish/nervousfish/test/LoginActivitySteps.java
jverbraeken: True, but this is an example file that will be deleted after a few weeks after we're accustomed to using Cucumber

app/build.gradle
jverbraeken: Because we have no directory called libs and because we shouldn't have one as we're using Gradle instead

PR #23, opened 2017-04-27
Implement Mockito

jverbraeken:
**Should be merged after #22** 

This PR implements Mockito

Why? Mockito makes it easier to test classes without having to rely on the implementation of other classes. Powermock is not recommended for Android development, because Powermock modifies the Android byte-code, but that doesn't work because of the Java Virtual Machine (Dalvik).

How? Mockito tests run exactly the same as plain Java tests. Note that tests that require an Android device are in the folder "androidTest" while plain Java tests are in the folder "test".

Feedback? Probably not needed ;-)


ericcornelissen:
I'll start reviewing this after #22 is merged 😉

PR #34, opened 2017-04-28
Add mockito support

jverbraeken:
This PR implements Mockito (and updates Cucumber to the newest version)

Why? Mockito makes it easier to test classes without having to rely on the implementation of other classes. Powermock is not recommended for Android development, because Powermock modifies the Android byte-code, but that doesn't work because of the Java Virtual Machine (Dalvik).

How? Mockito tests run exactly the same as plain Java tests. Note that tests that require an Android device are in the folder "androidTest" while plain Java tests are in the folder "test".

Feedback? Probably not needed ;-)


clenrock:
I agree with Eric, but if you are sure it works it's ok!

jverbraeken:
@ericcornelissen @clenrock Sample test added!

config/quality.gradle
```diff
@@ -52,6 +52,7 @@ task pmd(type: Pmd) {
     source 'src'
     include '**/*.java'
     exclude '**/gen/**'
+    exclude '**/test/**'
```
ericcornelissen: Two separate exclude statements seems a bit weird to me? 😕

jverbraeken: Yeah, but I can't find another way to do it

ericcornelissen: Well, I guess it is OK for now 👍

Issue #26, opened 2017-04-28
Loginscreen


jverbraeken:
As a user I want to have a login screen so that only I can access my public keys.

PR #39, opened 2017-04-30
Add the login screen

clenrock:
- Relevant issues: #26 

* * *

## What
With this pull request a working login screen is added to the app. 

## Why
In the future we want to secure the app with a password. It could also be that we use the password to encrypt the private key, but this still has to be discussed.

## How
The login screen is the new first screen you see when you open the app. The password is now hardcoded and is 12345. If you write something else, it will give an error. If you leave it empty you can still continue, for development purposes.

### DIFF
--- a/app/src/androidTest/assets/features/test.feature
+++ b/app/src/androidTest/assets/features/test.feature
@@ -1,8 +1,8 @@
 Feature: Login
-  Perform login on email and password are inputted
+  Perform login when a password is entered
 
 
-  Scenario Outline: Input email and password in correct format
+  Scenario Outline: Input and validate password
     Given I have a LoginActivity
     When I input password "<password>"
     And I press submit button
--- a/app/src/androidTest/assets/features/test.feature
+++ b/app/src/androidTest/assets/features/test.feature
@@ -1,15 +1,25 @@
 Feature: Login
   Perform login when a password is entered
 
-
-  Scenario Outline: Input and validate password
+  Scenario Outline: Input and validate password next activity
     Given I have a LoginActivity
     When I input password "<password>"
     And I press submit button
-    Then I should <see> auth error
+    Then I <should> continue to the MainActivity
 
     Examples:
-      | password   | see   |
-      | 85685 | true  |
-      | 12345  | false |
-      | 23235  | true  |
+      | password  | should |
+      | 59505  | false |
+      | 12345  | true |
+      | 23235  | false |
+
+  Scenario Outline: Input and validate wrong password
+      Given I have a LoginActivity
+      When I input password "<password>"
+      And I press submit button
+      Then I should see an auth error
+
+      Examples:
+        | password  |
+        | 75757  |
+        | 30572  |
--- a/app/src/androidTest/assets/features/test.feature
+++ b/app/src/androidTest/assets/features/test.feature
@@ -1,28 +1,15 @@
 Feature: Login
   Perform login on email and password are inputted
 
-  Scenario Outline: Input email and password in wrong format
-    Given I have a LoginActivity
-    When I input email <email>
-    And I input password "<password>"
-    And I press submit button
-    Then I should see error on the <view>
-
-    Examples:
-      | email | password  | view  |
-      | test  | lemoncake | email |
-      | test@test.com || password |
-
 
   Scenario Outline: Input email and password in correct format
     Given I have a LoginActivity
-    When I input email <email>
-    And I input password "<password>"
+    When I input password "<password>"
     And I press submit button
     Then I should <see> auth error
 
     Examples:
-      | email              | password   | see   |
-      | espresso@spoon.com | bananacake | true  |
-      | espresso@spoon.com | lemoncake  | false |
-      | latte@spoon.com    | lemoncake  | true  |
+      | password   | see   |
+      | 85685 | true  |
+      | 12345  | false |
+      | 23235  | true  |
--- a/app/src/androidTest/assets/features/test.feature
+++ b/app/src/androidTest/assets/features/test.feature
@@ -1,7 +1,7 @@
 Feature: Login
   Perform login when a password is entered
 
-  Scenario Outline: Input and validate password next activity
+  Scenario Outline: Input password and go to MainActivity or not
     Given I have a LoginActivity
     When I input password "<password>"
     And I press submit button
@@ -13,7 +13,7 @@ Feature: Login
       | 12345  | true |
       | 23235  | false |
 
-  Scenario Outline: Input and validate wrong password
+  Scenario Outline: Check that wrong password gives error
       Given I have a LoginActivity
       When I input password "<password>"
       And I press submit button
--- a/app/src/androidTest/java/com/nervousfish/nervousfish/test/LoginActivitySteps.java
+++ b/app/src/androidTest/java/com/nervousfish/nervousfish/test/LoginActivitySteps.java
@@ -24,7 +24,6 @@ import static android.support.test.espresso.action.ViewActions.typeText;
 import static android.support.test.espresso.assertion.ViewAssertions.matches;
 import static android.support.test.espresso.matcher.ViewMatchers.isDisplayed;
 import static android.support.test.espresso.matcher.ViewMatchers.withId;
-import static org.hamcrest.Matchers.not;
 
 @SuppressWarnings("PMD")
 @CucumberOptions(features = "features")
@@ -53,9 +52,9 @@ public class LoginActivitySteps extends ActivityInstrumentationTestCase2<LoginAc
     public void I_should_see_auth_error(boolean shouldSeeError) {
         if (shouldSeeError) {
             onView(withId(R.id.error)).check(matches(isDisplayed()));
-        } else {
-            onView(withId(R.id.error)).check(matches(not(isDisplayed())));
         }
+        //There is no else, because if there is no error it goes to the next activity, which
+        //doesn't have the error text.
     }
 
     private static Matcher<? super View> hasErrorText(String expectedError) {
--- a/app/src/androidTest/java/com/nervousfish/nervousfish/test/LoginActivitySteps.java
+++ b/app/src/androidTest/java/com/nervousfish/nervousfish/test/LoginActivitySteps.java
@@ -26,7 +26,6 @@ import static android.support.test.espresso.Espresso.onView;
 import static android.support.test.espresso.action.ViewActions.click;
 import static android.support.test.espresso.action.ViewActions.scrollTo;
 import static android.support.test.espresso.action.ViewActions.typeText;
-import static android.support.test.espresso.assertion.ViewAssertions.doesNotExist;
 import static android.support.test.espresso.assertion.ViewAssertions.matches;
 import static android.support.test.espresso.matcher.ViewMatchers.isDisplayed;
 import static android.support.test.espresso.matcher.ViewMatchers.withId;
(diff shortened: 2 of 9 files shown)


app/src/androidTest/assets/features/test.feature
```diff
@@ -1,28 +1,15 @@
...
-    And I press submit button
-    Then I should see error on the <view>
-
-    Examples:
-      | email | password  | view  |
-      | test  | lemoncake | email |
-      | test@test.com || password |
-
 
   Scenario Outline: Input email and password in correct format
```
ericcornelissen: Can you update the scenario outline as well?

app/src/androidTest/java/com/nervousfish/nervousfish/test/LoginActivitySteps.java
```diff
@@ -54,21 +48,13 @@ public void I_press_submit_button() {
...
-
     @Then("^I should (true|false) auth error$")
     public void I_should_see_auth_error(boolean shouldSeeError) {
         if (shouldSeeError) {
             onView(withId(R.id.error)).check(matches(isDisplayed()));
-        } else {
-            onView(withId(R.id.error)).check(matches(not(isDisplayed())));
         }
+        //There is no else, because if there is no error it goes to the next activity, which
+        //doesn't have the error text.
```
ericcornelissen: Isn't that the same as checking that the auth error is not displayed?

app/src/main/java/com/nervousfish/nervousfish/LoginActivity.java
```diff
@@ -12,58 +13,50 @@
...
 
-    private static final String DEMO_EMAIL = "espresso@spoon.com";
-    private static final String DEMO_PASSWORD = "lemoncake";
+    private static final String DEMO_PASSWORD = "12345";
 
-    private EditText mEmail, mPassword;
+    private EditText mPassword;
     private View mError;
 
+    private boolean inDevelopment = true;
```
ericcornelissen: Could you remove this? It is not really needed at this point in time... If you want to make sure we don't forget about the ability to login when no password is provided you can create an issue

The reason I don't want a variable like this is because if you want to be able to run the app in development or production mode this should be a global/environment variable and not a local (private) variable in a random class.

app/src/main/java/com/nervousfish/nervousfish/LoginActivity.java
```diff
@@ -12,58 +13,50 @@
...
-    }
-
     private void validateAccount() {
         if (mError == null) {
             mError = findViewById(R.id.error);
         }
 
-        if (!mEmail.getText().toString().equals(DEMO_EMAIL) || !mPassword.getText().toString().equals(DEMO_PASSWORD)) {
+        //The code below has to be changed before deployment. For development
+        //you can continue when the password is empty.
```
ericcornelissen: You can remove this comment if you create an issue for this, otherwise please change it to a TODO comment (`// TODO: Lorem ipsum dolor`)

app/src/main/java/com/nervousfish/nervousfish/LoginActivity.java
```diff
@@ -12,58 +13,50 @@
...
+        boolean wrongPassword = !mPassword.getText().toString().equals(DEMO_PASSWORD);
+        if(inDevelopment) {
```
ericcornelissen: We should be consistent with our spacing, please change this to `if (inDevelopment) {`. (It's wierd none of our development tools is catching this, I'll look into that 🤔)

app/src/androidTest/assets/features/test.feature
clenrock: Yes, thanks for pointing out

app/src/androidTest/java/com/nervousfish/nervousfish/test/LoginActivitySteps.java
clenrock: I put that there because maybe others would get confused that the code doesn't do anything. I'll remove it.

app/src/main/java/com/nervousfish/nervousfish/LoginActivity.java
clenrock: Good explaination, it's removed

app/src/main/java/com/nervousfish/nervousfish/LoginActivity.java
clenrock: Sure!

app/src/main/java/com/nervousfish/nervousfish/LoginActivity.java
clenrock: Checkstyle should have caught that! There will probably be some more errors

app/src/androidTest/java/com/nervousfish/nervousfish/test/LoginActivitySteps.java
ericcornelissen: Why not keep the `else` branch? As long as the description for the test method is `^I should (true|false) auth error$` it should be able to handle both cases...

app/src/main/java/com/nervousfish/nervousfish/LoginActivity.java
ericcornelissen: You know what, just add an issue for it anyway 😅

app/src/main/java/com/nervousfish/nervousfish/LoginActivity.java
```diff
@@ -24,46 +22,33 @@ protected void onCreate(Bundle savedInstanceState) {
...
-    }
-
     private void validateAccount() {
         if (mError == null) {
             mError = findViewById(R.id.error);
         }
 
-        if (!mEmail.getText().toString().equals(DEMO_EMAIL) || !mPassword.getText().toString().equals(DEMO_PASSWORD)) {
+        // TODO: have a real password check
+        boolean wrongPassword = !mPassword.getText().toString().isEmpty();
```
ericcornelissen: `wrongPassword` seems like a wierd name for this variable to me, given the value you're assigning to it... Right now the password is wrong if the string is not empty?

How about:
```java
boolean skipPassword = mPassword.getText().toString().isEmpty();
if (skipPassword == true) {
  // Next activity
} else {
  // 1. Check password
  // 2. Next activity (if the password is correct)
}
```

clenrock: That is of course the part we would change when we will deploy the app. But what you say makes it a bit clearer what happens ;)

app/src/androidTest/java/com/nervousfish/nervousfish/test/LoginActivitySteps.java
```diff
@@ -35,39 +34,24 @@ public LoginActivitySteps(LoginActivity activityClass) {
     }
 
     @Given("^I have a LoginActivity")
-    public void I_have_a_LoginActivity() {
+    public void i_have_a_LoginActivity() {
```
ericcornelissen: Can you change these method names so they're consistent with Java conventions? So this would become `iHaveALogInActivity()`

app/src/androidTest/java/com/nervousfish/nervousfish/test/LoginActivitySteps.java
```diff
@@ -35,39 +34,24 @@ public LoginActivitySteps(LoginActivity activityClass) {
...
-        int messageId = (viewName.equals("email")) ? R.string.msg_email_error : R.string.msg_password_error;
-
-        onView(withId(viewId)).check(matches(hasErrorText(getActivity().getString(messageId))));
-    }
-
     @Then("^I should (true|false) auth error$")
     public void I_should_see_auth_error(boolean shouldSeeError) {
         if (shouldSeeError) {
             onView(withId(R.id.error)).check(matches(isDisplayed()));
-        } else {
```
ericcornelissen: Why not keep the else branch? As long as the description for the test method is `^I should (true|false) auth error$` it should be able to handle both cases...

app/src/androidTest/java/com/nervousfish/nervousfish/test/LoginActivitySteps.java
clenrock: Good point, but the checkstyle checks want me to use two small letters to start with in a method name. I will check if I can change this checkstyle rule.

app/src/androidTest/java/com/nervousfish/nervousfish/test/LoginActivitySteps.java
clenrock: Because it now goes to another activity, a check I could use is to check that I actually went to that activity. For that I need to keep track of which activity we are in, we discussed this and you told me that it was bad practice to create something only because you want to use it for testing. Do you still think this is true, or is there some other way to test it?

ericcornelissen: I still think that is true, but that doesn't apply here.

This test is named "I_should_see_auth_error" so that is wat it should test, this is not at all related to switching to a different activity. So:
- If you're using this test to see if the activity changes, you're using it incorrectly and you should change that. Or,
- if you're not using this test for that, you should still keep the `else` branch because, as I stated before, the description/annotation of the test still describes a situation where the else branch is needed.

app/src/androidTest/java/com/nervousfish/nervousfish/test/LoginActivitySteps.java
ericcornelissen: In that case we'll have to update the Checkstyle ruleset :wink: You can find more info about the relevant rule [here](http://checkstyle.sourceforge.net/config_naming.html#MethodName)

app/src/androidTest/java/com/nervousfish/nervousfish/test/LoginActivitySteps.java
clenrock: Fixed it by checking if it exists or not.

ericcornelissen: Why not leave the test as it was ?

app/src/androidTest/assets/features/test.feature
```diff
@@ -1,28 +1,25 @@
 Feature: Login
-  Perform login on email and password are inputted
+  Perform login when a password is entered
 
-  Scenario Outline: Input email and password in wrong format
+  Scenario Outline: Input and validate password next activity
```
ericcornelissen: Again, the Scenario Outline should be a well-structured English sentence, e.g. "Input correct passwords and go to the main activity"

app/src/androidTest/assets/features/test.feature
```diff
@@ -1,28 +1,25 @@
...
     Examples:
-      | email | password  | view  |
-      | test  | lemoncake | email |
-      | test@test.com || password |
+      | password  | should |
+      | 59505  | false |
+      | 12345  | true |
+      | 23235  | false |
 
+  Scenario Outline: Input and validate wrong password
```
ericcornelissen: Again, the Scenario Outline should be a well-structured English sentence.

app/src/androidTest/java/com/nervousfish/nervousfish/test/LoginActivitySteps.java
```diff
@@ -35,40 +39,48 @@ public LoginActivitySteps(LoginActivity activityClass) {
...
+        onView(withId(R.id.error)).check(matches(isDisplayed()));
     }
 
-    @Then("^I should (true|false) auth error$")
-    public void I_should_see_auth_error(boolean shouldSeeError) {
-        if (shouldSeeError) {
-            onView(withId(R.id.error)).check(matches(isDisplayed()));
-        } else {
-            onView(withId(R.id.error)).check(matches(not(isDisplayed())));
+    Activity getCurrentActivity() {
```
ericcornelissen: Can you make this a `private` method please. Also, I'm not quite sure how this is supposed to work... Does [line 43](https://github.com/ericcornelissen/NervousFish/blob/96acc8e91c2b4a817f4d83a09a84eecddba4b7e8/app/src/androidTest/java/com/nervousfish/nervousfish/test/LoginActivitySteps.java#L43) reference this method?

clenrock: Good point, and no that points to the object it extends. It returns the activity which it is initialized with at the constructor.

jverbraeken:
@clenrock Merge conflicts...

Issue #32, opened 2017-04-28
Project structure


jverbraeken:
As a developer I want to have an overview (eg UML) of the project so that I can start working on my class without having to wait for the classes it depends on to be finished.

jverbraeken:
Duplicate

ericcornelissen:
We're not yet done with this, we have to hand in the final version next week 😉

PR #41, opened 2017-05-01
Structure

jverbraeken:
- Relevant Issues: #32 
- Related Pull Requests: -

* * *

## What
This PR add the main structure to the project.

## Why
This is done so that all developers can start coding against an interface that provides a rough direction and to make sure that the project is extensible, safe and fast.

## How
3 things are central: encapsulation (everything private), immutability (every final) and extensibility (everything what could be extended in the future coded against an interface).
All functionality is placed in modules to have the project modular. To give classes access to these modules they can a certain class, namely a Service Locator, for any of these modules.
Singletons are bad, so therefore the Service Locator is dependency-injected (so passed as a parameter) to new classes that may need the service locator. To get the highest degree of encapsulation possible, the service locator constructor is (package-)private. To construct a new ServiceLocator the client can use a new ServiceLocatorCreator. His constructor is also (package-)private so this class can only be made by MainActivity (the only "system" Activity). All modules have private constructors as well and the service locator can request modules by asking the class to wrap a new instance of the class in a ModuleWrapper. The reason that this is done is to make sure that all variables are final (even the circular dependency between the service locator and every module) and to make sure that all modules of a service locator are initialized.
A few class names are pretty long, but that's ok because the client will only use the interface, the class itself is irrelevant is only used once in the ServiceLocatorCreator.

## Notes
I opened this PR without running any of the code quality tools because they do not work for me yet and this pr is needed. So running the code quality tools while reviewing is appreciated.


app/src/main/java/com/nervousfish/nervousfish/modules/IModule.java
```diff
@@ -0,0 +1,8 @@
+package com.nervousfish.nervousfish.modules;
+
+/**
+ * Defines a module that can be used by a service locator.
```
ericcornelissen: You can use `@see com.nervousfish.nervousfish.ServiceLocator` to reference the class 😉

app/src/main/java/com/nervousfish/nervousfish/modules/constants/IConstants.java
```diff
@@ -0,0 +1,9 @@
+package com.nervousfish.nervousfish.modules.constants;
+
+import com.nervousfish.nervousfish.modules.IModule;
+
+/**
+ * Interface for the Constants classes
```
ericcornelissen: Should end with a period and can be more descriptive

app/src/main/java/com/nervousfish/nervousfish/modules/cryptography/IKeyGenerator.java
```diff
@@ -0,0 +1,10 @@
+package com.nervousfish.nervousfish.modules.cryptography;
+
+import com.nervousfish.nervousfish.modules.IModule;
+
+/**
+ * Defines a module for generating secure keys that can be used by a service locator.
+ */
+
+public interface IKeyGenerator  extends IModule {
```
ericcornelissen: One space to much in between `IKeyGenerator` and `extends`

app/src/main/java/com/nervousfish/nervousfish/service_locator/IServiceLocatorCreator.java
```diff
@@ -0,0 +1,9 @@
+package com.nervousfish.nervousfish.service_locator;
+
+
+/**
+ * Defines the interface of a class that can create a new {@link IServiceLocator}.
+ */
+public interface IServiceLocatorCreator {
```
ericcornelissen: Do we really need a dedicated class for this?

app/src/main/java/com/nervousfish/nervousfish/modules/IModule.java
```diff
@@ -0,0 +1,8 @@
...
+ */
+
+public interface IModule {
```
ericcornelissen: Why do we need this interface ?

jverbraeken: It's used in ModuleWrapper<T extends IModule>

app/src/main/java/com/nervousfish/nervousfish/service_locator/IServiceLocatorCreator.java
jverbraeken: It's possible that we have multiple Service Locators like we had in SEM. The modules have a reference to a service locator, so we'll have to use an interface to provide support for multiple service locators.

jverbraeken:
@ericcornelissen I've implemented all your suggestions!

app/src/main/java/com/nervousfish/nervousfish/service_locator/IServiceLocatorCreator.java
ericcornelissen: Then why is it called a "creator" ?

jverbraeken: Because every service locator needs its own "creator", otherwise the serviceLocator variable cannot be final

ericcornelissen: 👍

app/src/main/java/com/nervousfish/nervousfish/modules/pairing/AndroidNFCAdapter.java
```diff
@@ -0,0 +1,29 @@
+package com.nervousfish.nervousfish.modules.pairing;
+
+import com.nervousfish.nervousfish.service_locator.IServiceLocator;
+import com.nervousfish.nervousfish.service_locator.IServiceLocatorCreator;
+import com.nervousfish.nervousfish.service_locator.ModuleWrapper;
+
+/**
+ * An adapter to the default Android bluetooth library
+ */
+public final class AndroidNFCAdapter implements INFCHandler {
```
clenrock: Is it useful to have the NFC adapter and the QR adapter in if we are not busy with those implementations yet?

app/src/main/java/com/nervousfish/nervousfish/modules/pairing/ZXingQRAdapter.java
```diff
@@ -0,0 +1,29 @@
+package com.nervousfish.nervousfish.modules.pairing;
+
+import com.nervousfish.nervousfish.service_locator.IServiceLocator;
+import com.nervousfish.nervousfish.service_locator.IServiceLocatorCreator;
+import com.nervousfish.nervousfish.service_locator.ModuleWrapper;
+
+/**
+ * An adapter to the ZXing QR code library
+ */
+public final class ZXingQRAdapter implements IQRHandler {
```
clenrock: Are you already sure we should use ZXing? I think this class also should not be here yet because we are not going to implement it at this moment

app/src/main/java/com/nervousfish/nervousfish/modules/pairing/AndroidNFCAdapter.java
jverbraeken: No,but the Service Locator must have a valid reference to an implementation of the pairing modules. To implement your suggestion I removed these classes and made 3 anonymous classes (from their interfaces) in ServiceLocatorCreator

app/src/main/java/com/nervousfish/nervousfish/modules/pairing/ZXingQRAdapter.java
jverbraeken: See above

ericcornelissen: I think ZXing QR code is to specific for now, we know we want to support QR codes but haven't discussed what QR codes yet, so I agree with @clenrock on this one 🙂

app/src/main/java/com/nervousfish/nervousfish/side_activities/LoginActivity.java
```diff
@@ -1,11 +1,13 @@
-package com.nervousfish.nervousfish;
+package com.nervousfish.nervousfish.side_activities;
```
ericcornelissen: Why don't we put all activities in the same package?

config/quality.gradle
```diff
@@ -15,6 +15,8 @@ task checkstyle(type: Checkstyle) {
     source 'src'
     include '**/*.java'
     exclude '**/gen/**'
+    exclude '**/androidTest/**'
+    exclude '**/test/**'
```
ericcornelissen: Why shouldn't the tests adhere to code quality standards?

app/src/main/java/com/nervousfish/nervousfish/side_activities/LoginActivity.java
jverbraeken: Because we only want a single Service Locator, made by the very first Activity created. To make sure that only that very first activity, namely MainActivity, can create a new one, we put the MainActivity in the same package as the ServiceLocator classes and make the constructor package-private. The other activities cannot create a second service locator anymore when they are in a seperate package and should get the service locator dependency injected by MainActivity

config/quality.gradle
jverbraeken: The cucumber tests have very, very long names and the unit tests can have a lot of methods. We can also suppress specific warnings in the tests if you like.

ericcornelissen: I'd most certainly prefer that 😄

app/src/main/java/com/nervousfish/nervousfish/side_activities/LoginActivity.java
ericcornelissen: Wouldn't it be possible to create a different entry class which can be in the same package as the ServiceLocator? I'm not sure how this works, but I imagine it is possible...

config/quality.gradle
TheBonheurs: Agree!

app/src/main/java/com/nervousfish/nervousfish/modules/constants/Constants.java
```diff
@@ -0,0 +1,30 @@
...
+        return new ModuleWrapper<>(new Constants(serviceLocatorCreator));
+    }
+
+    /**
+     * Prevents construction from outside the class.
+     * @param serviceLocatorCreator The object responsible for creating the service locator
+     */
+    @SuppressWarnings("PMD.UnusedFormalParameter")
+    private Constants(final IServiceLocatorCreator serviceLocatorCreator) {
+        // final IServiceLocator serviceLocator = serviceLocatorBridge.getServiceLocator();
```
ericcornelissen: Why is this not implemented yet? Isn't it possible to do something like: (also applies to other similar instances)
```java
this.serviceLocator = serviceLocatorCreate.getServiceLocator();
```

app/src/main/java/com/nervousfish/nervousfish/service_locator/ModuleWrapper.java
```diff
@@ -0,0 +1,20 @@
...
+ */
+
+public final class ModuleWrapper<T extends IModule> {
+    private final T module;
+
+    public ModuleWrapper(final T module) {
+        this.module = module;
+    }
+
+    T get() {
```
ericcornelissen: Is this supposed to be protected? If so consider adding `/* protected */`

app/src/main/java/com/nervousfish/nervousfish/ConstantKeywords.java
```diff
@@ -0,0 +1,15 @@
+package com.nervousfish.nervousfish;
+
+/**
+ * Contains the keywords used in the application.
+ */
+public final class ConstantKeywords {
+    public static final String SERVICE_LOCATOR = "a";
```
jverbraeken: In EntryActivity we pass the service locator to the mainactivity, that must be identified by a certain string. By using a constant String we won't have to worry about typing errors

app/src/main/java/com/nervousfish/nervousfish/modules/constants/Constants.java
jverbraeken: I commented it out to remove the PMD warning, but still give you the code of how you can get the service locator when you need it

app/src/main/java/com/nervousfish/nervousfish/service_locator/ModuleWrapper.java
jverbraeken: It's package-private ;-)

app/src/main/java/com/nervousfish/nervousfish/ConstantKeywords.java
ericcornelissen: Okay, in that case would you mind changing the string to "service locator" 😇

.idea/modules.xml
```diff
@@ -3,6 +3,8 @@
   <component name="ProjectModuleManager">
     <modules>
       <module fileurl="file://$PROJECT_DIR$/NervousFish.iml" filepath="$PROJECT_DIR$/NervousFish.iml" />
+      <module fileurl="file://$PROJECT_DIR$/NervousFish.iml" filepath="$PROJECT_DIR$/NervousFish.iml" />
+      <module fileurl="file://$PROJECT_DIR$/app/app.iml" filepath="$PROJECT_DIR$/app/app.iml" />
       <module fileurl="file://$PROJECT_DIR$/app/app.iml" filepath="$PROJECT_DIR$/app/app.iml" />
```
ericcornelissen: Could you revert these changes please 🙂

app/src/main/java/com/nervousfish/nervousfish/activities/LoginActivity.java
```diff
@@ -7,30 +7,39 @@
 import android.view.View;
 import android.widget.EditText;
 
+import com.nervousfish.nervousfish.R;
+
 /**
  * Demo Login activity class
  */
 @SuppressWarnings("PMD")
```
ericcornelissen: Why are you suppressing PMD? (if this is because the class isn't fully implemented yet, add a comment explaining that and at what point the suppression can be removed)

app/src/main/java/com/nervousfish/nervousfish/modules/pairing/APairingHandler.java
```diff
@@ -0,0 +1,15 @@
...
+ */
+@SuppressWarnings("PMD.AbstractClassWithoutAbstractMethod")
+abstract class APairingHandler {
+
+    /**
+     * Prevent instatiation from non-subclasses outside the package
+     */
+    @SuppressWarnings("PMD.UncommentedEmptyConstructor")
+    protected APairingHandler() {
+    }
```
ericcornelissen: Consider removing the suppression, adding the JavaDoc as a comment inside the method and changing the JavaDoc to "Simple {@code APairingHandler} constructor without parameters."

config/quality/checkstyle/checkstyle.xml
```diff
@@ -1,140 +1,238 @@
-<?xml version="1.0"?>
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
ericcornelissen: This shouldn't be in this PR...

.idea/modules.xml
ericcornelissen: @jverbraeken I asked you to revert the changes, not remove the file 😅

Issue #28, opened 2017-04-28
Generate private / public key


jverbraeken:
As a user I want to have a private and public key so that I can communicate with others encrypted

- [ ] Generate private / public key
- [ ] Store the keys in the database

PR #47, opened 2017-05-03
App/generate key pair

clenrock:
Relevant Issues: #28

* * *

NOTE; this pull request is now opened only for review. When the pull request #41 is merged in develop I will finish this. Also only check the class KeyPairGenerator, the other interfaces are of no use now.

## What
This pull request adds a way to create a public/private Key Pairs.

## Why
When people want to exchange keys, they need to have generated one first.


app/src/main/java/com/nervousfish/nervousfish/cryptography/KeyGenerator.java
```diff
@@ -0,0 +1,62 @@
...
+
+import java.security.KeyFactory;
+import java.security.KeyPair;
+import java.security.KeyPairGenerator;
+import java.security.NoSuchAlgorithmException;
+import java.security.spec.InvalidKeySpecException;
+import java.security.spec.RSAPrivateKeySpec;
+import java.security.spec.RSAPublicKeySpec;
+
+public class KeyGenerator {
```
ericcornelissen: @jverbraeken do we require class JavaDoc for classes, if so we should add a Checkstyle rule for that and add it here...

app/src/main/java/com/nervousfish/nervousfish/cryptography/KeyGenerator.java
```diff
@@ -0,0 +1,62 @@
...
+            return fact.getKeySpec(kp.getPublic(),
+                    RSAPublicKeySpec.class);
+        } catch (NoSuchAlgorithmException | InvalidKeySpecException e) {
+            e.printStackTrace();
+        }
+        return null;
+    }
+
+    /**
+     * Gets from kp the private key and makes a RSAPrivateKeySpec out of it.
```
ericcornelissen: `Gets from kp...`? I assume this reference to the application? The documentation for a method should not specify *how* it generates the keypair but that it does that (this applies to the documentation of the other methods as well)?

ericcornelissen:
@jverbraeken @clenrock 

Currently the keypair generator methods ([1](https://github.com/ericcornelissen/NervousFish/blob/9c890e827764ae7b4c1dc1884159c920cae69657/app/src/main/java/com/nervousfish/nervousfish/cryptography/KeyGenerator.java#L17), [2](https://github.com/ericcornelissen/NervousFish/blob/9c890e827764ae7b4c1dc1884159c920cae69657/app/src/main/java/com/nervousfish/nervousfish/cryptography/KeyGenerator.java#L34), [3](https://github.com/ericcornelissen/NervousFish/blob/9c890e827764ae7b4c1dc1884159c920cae69657/app/src/main/java/com/nervousfish/nervousfish/cryptography/KeyGenerator.java#L51)) all catch errors internally and return `null` if the key could not be generated. But I think this is not the way to go.

A better solution, in my opinion, would be to delegate these errors to the caller of these methods. This because the error is more informative then just a `null` value to the caller, and they have to handle the error anyway (also making it more intuitive to use the method). The only effect of this change (I think) would be that the following code (where a keypair is generated):
```java
KeyGenerator keyGenerator = new KeyGenerator();
KeyPair keyPair = keyGenerator.generateRandomKeyPair();
if (keyPair == null) {
  // inform user about error
} else {
  // do useful stuff
}
```

Would change into:
```java
KeyGenerator keyGenerator = new KeyGenerator();
try {
   KeyPair keyPair = keyGenerator.generateRandomKeyPair();
  // do usefull stuff
} catch (...) {
  // inform user about error
}
```

What are your thoughts on this? 😁

app/src/main/java/com/nervousfish/nervousfish/cryptography/KeyGenerator.java
clenrock: 'kp' is the key pair which is used as parameter. Maybe I should make it more clear that that is an parameter.

ericcornelissen: Ah, you can use something like `{@code foo}` to make it at least little bit more clear 😉 Also, you might consider renaming the parameter to `keyPair`.

clenrock:
@ericcornelissen I think I prefer your suggestion, but it doesn't matter a lot ;). If @jverbraeken also thinks your suggestion is better I'll change it.

app/src/main/java/com/nervousfish/nervousfish/cryptography/KeyGenerator.java
jverbraeken: That file doesn't exist anymore, but I changed CheckStyle so that it errors on lack of documentation

jverbraeken: Done, in structure...

app/src/main/java/com/nervousfish/nervousfish/cryptography/KeyGenerator.java
```diff
@@ -0,0 +1,62 @@
...
+    public static KeyPair generateRandomKeyPair() {
+        try {
+            KeyPairGenerator k = KeyPairGenerator.getInstance("RSA");
+
+            k.initialize(2048);
+            return k.generateKeyPair();
+        } catch (NoSuchAlgorithmException e) {
+            e.printStackTrace();
+        }
+        return null;
```
jverbraeken: I think it's better to throw a new Exception("explanation of what went wrong") than to return null, because now the program just continues and we may not even notice that something went wrong

app/src/main/java/com/nervousfish/nervousfish/cryptography/KeyGenerator.java
```diff
@@ -0,0 +1,62 @@
...
+    public static RSAPublicKeySpec getPublicKeySpec(KeyPair kp) {
+        try {
+            KeyFactory fact = KeyFactory.getInstance("RSA");
+
+            return fact.getKeySpec(kp.getPublic(),
+                    RSAPublicKeySpec.class);
+        } catch (NoSuchAlgorithmException | InvalidKeySpecException e) {
+            e.printStackTrace();
+        }
+        return null;
```
jverbraeken: same

app/src/main/java/com/nervousfish/nervousfish/cryptography/KeyGenerator.java
```diff
@@ -0,0 +1,62 @@
...
+
+public class KeyGenerator {
+
+    /**
+     * Generates a random KeyPair with the RSA algorithm.
+     * @return a randomly generated KeyPair
+     */
+    public static KeyPair generateRandomKeyPair() {
+        try {
+            KeyPairGenerator k = KeyPairGenerator.getInstance("RSA");
```
jverbraeken: It should be final, otherwise it will error on the new Checkstyle rules (and also because it's good coding style because it's easier to reason about the program)

app/src/main/java/com/nervousfish/nervousfish/cryptography/KeyGenerator.java
```diff
@@ -0,0 +1,62 @@
...
+import java.security.spec.RSAPrivateKeySpec;
+import java.security.spec.RSAPublicKeySpec;
+
+public class KeyGenerator {
+
+    /**
+     * Generates a random KeyPair with the RSA algorithm.
+     * @return a randomly generated KeyPair
+     */
+    public static KeyPair generateRandomKeyPair() {
```
jverbraeken: After structure is merged, these methods should not be static, but just normal methods

app/src/androidTest/assets/features/publicPrivateKeyGeneration.feature
```diff
@@ -0,0 +1,15 @@
+Feature: Public and private key generation
+  As a user
+  I want to have the app generate a public/private key pair when I use it for the first time
+  So that I don't have to generate a pair myself
```
TheBonheurs: Aren't these pairs also used later to establish a secure connection between two users. What I mean to say is maybe this is not the only reason.

app/src/main/java/com/nervousfish/nervousfish/cryptography/KeyGenerator.java
```diff
@@ -0,0 +1,62 @@
...
+
+    /**
+     * Gets from kp the private key and makes a RSAPrivateKeySpec out of it.
+     * @param kp the KeyPair with the private key
+     * @return the RSAPrivateKeySpec
+     */
+    public static RSAPrivateKeySpec getPrivateKeySpec(KeyPair kp) {
+        try {
+            KeyFactory fact = KeyFactory.getInstance("RSA");
+
```
TheBonheurs: Why the newline here?

app/src/androidTest/assets/features/publicPrivateKeyGeneration.feature
ericcornelissen: No they aren't, we are not concerned with encrypting the Bluetooth/NFC/whatever connection

app/src/main/java/com/nervousfish/nervousfish/cryptography/KeyGenerator.java
ericcornelissen: 👍

app/src/androidTest/assets/features/publicPrivateKeyGeneration.feature
TheBonheurs: But when confirming the other user on an established connection aren't we using the assistance of the pairs in our extra protection layer?

ericcornelissen: The idea is that the confirmation is done through user input (think something like the Challenge-Response model). Encryption is not needed here for three reasons:
- The connection itself is already secure. Bluetooth has build-in encryption, NFC is very short range so interception is nearly impossible (other mediums are, admittedly, possibly less secure).
- What the confirmation is trying to accomplish is a guarantee that I actually received your key and not someone else's key.
- The public keys are public after all, it is not that bad if they get out in the open world... All I really care about is that I really have your key and I can really send secure message to you specifically.

TheBonheurs: > What the confirmation is trying to accomplish is a guarantee that I actually received your key and not someone else's key.
So we are exchanging them, I see that as a use of a key. Maybe that's where our definitions don't comply. Anyways this is a small detail to keep worrying over, keep this as you like @clenrock

ericcornelissen: I think our definitions don't comply indeed 😛 We don't use the key when it is exchanged, we could exchange whatever we want (for that matter, we will probably exchange the name of the user as well).

Think about it like this, we're trying to exchange random strings of text. These strings just happen to be public keys...

app/src/main/java/com/nervousfish/nervousfish/modules/cryptography/KeyGeneratorAdapter.java
```diff
@@ -28,4 +36,41 @@ private KeyGeneratorAdapter(final IServiceLocatorCreator serviceLocatorCreator)
     public static ModuleWrapper<KeyGeneratorAdapter> newInstance(final IServiceLocatorCreator serviceLocatorCreator) {
         return new ModuleWrapper<>(new KeyGeneratorAdapter(serviceLocatorCreator));
     }
+
+    /**
+     * Generates a random KeyPair with the RSA algorithm.
+     * @return a randomly generated KeyPair
+     */
+    public static KeyPair generateRandomKeyPair() throws NoSuchAlgorithmException {
```
ericcornelissen: I don't really like this method name... I would suggest changing it to `generateRSAKeyPair()`, removing the random key pair as that is not really relevant and adding the key pair type so you don't have to go and read the documentation in order to know what key you're generating.

Having said that, I would update the documentation to something like: 
```java
/**
 * Generate a RSA {@link KeyPair}.
 *
 * @return A randomly generated {@link KeyPair}
 */
```

app/src/main/java/com/nervousfish/nervousfish/modules/cryptography/KeyGeneratorAdapter.java
```diff
@@ -28,4 +36,41 @@ private KeyGeneratorAdapter(final IServiceLocatorCreator serviceLocatorCreator)
...
+
+        return k.generateKeyPair();
+    }
+
+    /**
+     * Gets from the parameter 'kp' the public key and makes a RSAPublicKeySpec out of it.
+     * @param keyPair the KeyPair with the public key
+     * @return the RSAPublicKeySpec
+     */
+    public static RSAPublicKeySpec getPublicKeySpec(final KeyPair keyPair) throws NoSuchAlgorithmException,
```
ericcornelissen: Are we interested in the `RSAPublicKeySpec` (and to an extend the `RSAPrivateKeySpec`)? Don't we just want to call the `generateRandomKeyPair()` and get a usable public/private KeyPair (probably our own implementation where the keys are directly accessible as strings)?

clenrock: We can make our own objects which have these as string. Maybe that is a bit nicer because the above code only is needed for RSA keys. Would be nice to have a bit more generic object.
Would the RSAKey POJO be all right?

app/src/main/java/com/nervousfish/nervousfish/modules/cryptography/KeyGeneratorAdapter.java
clenrock: Good point, I will change that

app/src/main/java/com/nervousfish/nervousfish/modules/cryptography/KeyGeneratorAdapter.java
ericcornelissen: I agree 100% 👍

app/src/main/java/com/nervousfish/nervousfish/data_objects/KeyPair.java
```diff
@@ -0,0 +1,22 @@
+package com.nervousfish.nervousfish.data_objects;
+
+/**
+ * KeyPair POJO which keeps a public and private key.
+ */
+
+public class KeyPair {
+
+    public IKey publicKey;
+    public IKey privateKey;
```
ericcornelissen: These can (in the context of a POJO) and should be `private` and `final`

app/src/main/java/com/nervousfish/nervousfish/data_objects/RSAKey.java
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * RSA variant of {@link IKey}.
+ */
+public final class RSAKey implements IKey {
+
+    private final static String TYPE = "RSA";
+
+    public final String exponent;
+    public final String modulus;
```
ericcornelissen: These can (in the context of a POJO) and should be `private`

app/src/main/java/com/nervousfish/nervousfish/data_objects/RSAKey.java
```diff
@@ -0,0 +1,63 @@
+package com.nervousfish.nervousfish.data_objects;
+
+/**
+ * RSA variant of {@link IKey}.
+ */
+public final class RSAKey implements IKey {
+
+    private final static String TYPE = "RSA";
```
ericcornelissen: The value should come from `ConstantKeywords`

app/src/main/java/com/nervousfish/nervousfish/data_objects/RSAKey.java
```diff
@@ -0,0 +1,63 @@
...
+    @Override
+    public String getType() {
+        return RSAKey.TYPE;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    @SuppressWarnings("PMD")
```
ericcornelissen: This should specify the rule it is suppressing

app/src/main/java/com/nervousfish/nervousfish/modules/cryptography/KeyGeneratorAdapter.java
```diff
@@ -41,36 +42,20 @@ private KeyGeneratorAdapter(final IServiceLocatorCreator serviceLocatorCreator)
      * Generates a random KeyPair with the RSA algorithm.
      * @return a randomly generated KeyPair
      */
-    public static KeyPair generateRandomKeyPair() throws NoSuchAlgorithmException {
+    public static KeyPair generateRSAKeyPair() throws NoSuchAlgorithmException, InvalidKeySpecException {
         final KeyPairGenerator k = KeyPairGenerator.getInstance("RSA");
```
ericcornelissen: Could you rename this variable to `key` for readability?

app/src/main/java/com/nervousfish/nervousfish/modules/cryptography/KeyGeneratorAdapter.java
```diff
@@ -41,36 +42,20 @@ private KeyGeneratorAdapter(final IServiceLocatorCreator serviceLocatorCreator)
      * Generates a random KeyPair with the RSA algorithm.
      * @return a randomly generated KeyPair
```
ericcornelissen: Could you change this to:
```java
/**
 * Generate a random key pair with the RSA algorithm.
 *
 * @return A new {@link KeyPair}.
 */
```

app/src/main/java/com/nervousfish/nervousfish/data_objects/KeyPair.java
clenrock: Fixed 😄

ericcornelissen: Are you sure? 😉

clenrock: Can't read final

clenrock: Shit, can't read both

app/src/main/java/com/nervousfish/nervousfish/data_objects/KeyPair.java
```diff
@@ -0,0 +1,38 @@
+package com.nervousfish.nervousfish.data_objects;
+
+/**
+ * KeyPair POJO which keeps a public and private key.
+ */
+
+public class KeyPair {
```
ericcornelissen: This class should be `final`

app/src/main/java/com/nervousfish/nervousfish/data_objects/KeyPair.java
```diff
@@ -0,0 +1,38 @@
...
+        this.publicKey = publicKey;
+        this.privateKey = privateKey;
+    }
+
+    /**
+     * Returns the public key of this KeyPair.
+     * @return the public key as IKey
+     */
+    public IKey getPublicKey() {
+        return publicKey;
```
ericcornelissen: Would you mind adding `this.` 😇

app/src/main/java/com/nervousfish/nervousfish/data_objects/KeyPair.java
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    /**
+     * Returns the private key of this KeyPair.
+     * @return the private key as IKey
+     */
+    public IKey getPrivateKey() {
+        return privateKey;
```
ericcornelissen: Would you mind adding `this.` 😇

app/src/main/java/com/nervousfish/nervousfish/modules/cryptography/KeyGeneratorAdapter.java
```diff
@@ -28,4 +37,25 @@ private KeyGeneratorAdapter(final IServiceLocatorCreator serviceLocatorCreator)
...
+    /**
+     * Generates a random KeyPair with the RSA algorithm.
+     * @return a randomly generated KeyPair
+     */
+    public static KeyPair generateRSAKeyPair() throws NoSuchAlgorithmException, InvalidKeySpecException {
+        final KeyPairGenerator k = KeyPairGenerator.getInstance("RSA");
+        k.initialize(2048);
+
+        final java.security.KeyPair keyPair = k.generateKeyPair();
+        final KeyFactory fact = KeyFactory.getInstance("RSA");
```
ericcornelissen: Could you use the variable names `key` instead of `k` and `keyFactory` instead of `fact`

PR #57, opened 2017-05-05
Contacts part of the database

ericcornelissen:
- Relevant Issues: #27
- Related Pull Requests: n/a

* * *

## What
With this pull request a GSON implementation of the database will be added which, for now, is capable of:
- [x] Create a new contact in the database (i.e. an entry with public key and name).
- [x] Modify a contact in the database (i.e. change the name).
- [x] Delete a contact from the database (i.e. remove the entry, key+name, completely).
- [x] Get a list of all contacts from the database (i.e. a list of the names of all contacts).
- [x] ~~Get a specific contact from the database (i.e. all information about that contact, name+key).~~

## Why
We want to store the contacts on the phone so they can be viewed/used at a later point in time

## How
The database adapter in the PR is implemented using Google's [gson](https://github.com/google/gson), as a result the database is implemented in JSON files.

## Alternative implementation
We have considered using other simple text-based database implementation (such as .csv) but felt that a JSON implementation is the right combination of simple and extensible.

## Notes
- The `ExampleUnitTest` has been removed as this is the first PR with actual unit tests.
- The class `KeyAdapter` in the `modules/database` package is needed to convert the arbitrary key representation in the database to an actual key POJO implementing the `IKey` interface.
- Each `IKey` implementing class needs a `getType()` method so the `KeyAdapter` can use the type to figure out what instance of `IKey` it must create when reading the database.
- This PR is somewhat related to #56 as there is an issue with the `ServiceLocator` (this also exists in develop), so there might be some issues but these will be fixed soon

Issue #27, opened 2017-04-28
Database


jverbraeken:
[//]: # "Check if the title is descriptive!"
- Parent issues: #1
- Relevant issues: #10, #11, #12, #14
- Estimated effort: 4 hours

* * *

## User story
As a developer I want to have a database in which the data of the user is stored

## What
We want to have a database in which user data is stored. The user data can be split up into two fraction: the user data (public/private key pair+name) and contact data (public keys+name of other people).

## Why
The data of the user is there so it can be used by other application and so the public key+name can be exchanged with others.

The data of the contacts is needed so the user can send encrypted messages to them using their public key. The name is needed so the user can identify them.

## How
There will be a database package that will be used to perform the action described above so the application can actually store data into/read from the database.

The actual database will be implemented in JSON files because it is easy and good enough (at least for now) since the data/queries are not complex and it is only needed locally.

## Definition of done
This task is done once it is possible to perform certain tasks with regards to the database:
- [x] Create a new contact in the database (i.e. an entry with public key and name).
- [x] Modify a contact in the database (i.e. change the name).
- [x] Delete a contact from the database (i.e. remove the entry, key+name, completely).
- [x] Get a list of all contacts from the database (i.e. a list of the names of all contacts).
- [x] ~~Get a specific contact from the database (i.e. all information about that contact, name+key).~~
- [x] Get your own details from the database (i.e. all information about you, name+key).
- [x] Create your info, public and private key+name, in the database.
- [x] Update your name in the database.

## Notes
n/a

ericcornelissen:
> Get a specific contact from the database (i.e. all information about that contact, name+key).

I propose we remove this bullet, as it is not relevant to our implementation. First and foremost because once we request the list of all contact, we already have all information about each contact so we don't need to request more information. Also, this would require some kind of identifier (requesting a contact using the `Contact` class would be a little silly) which we don't really need...

What do you think @clenrock?

clenrock:
@ericcornelissen I agree! We got everything in an array already

ericcornelissen:
I updated the description 👍

Issue #50, opened 2017-05-03
Have the possibility to have multiple accounts


clenrock:
- Parent issues: #27 
- Estimated effort: 5

* * *

## User story
As a user I want to be able to use multiple accounts, so that I can send certain people one public key while I can send another to other people.

## What
This adds an option to create multiple accounts and add these to the database. There are 2 ways we can implement this. The first one being 2 totally different accounts, with their separate contacts list. And the other being one account per app, but the choice of which public key to send before connecting to someone.

## Why
This resolves the problem that people who use multiple public keys, at for example different computers, will also need 2 phones and two different apps to keep their key on.

## How
The two main parts that need to be changed is in the database the ability to add new accounts to the account information file. In the app itself it should be possible to create a new account, and to select this account.

## Definition of done
- [ ] Give the ability in the app to create an additional account
- [ ] Modify the database to handle multiple accounts
- [ ] Modify the database to easily add a new account to the list
- [ ] Give the ability in the app to choose a certain account

## Notes
We need to make a decision about the problem described in the 'How' paragraph.

ericcornelissen:
Seems like a good idea to me! 👍 I've got some questions tho.

First and foremost, I would suggest to change the name from "Accounts" to "Profiles", but I'm open to other suggestions (accounts just reminds me too much of a thing where you have to login...).

* * *

> There are 2 ways we can implement this. The first one being 2 totally different accounts, with their separate contacts list. And the other being one account per app, but the choice of which public key to send before connecting to someone.

I dislike the first one very much, apart from the fact that a system like this could be useful when the app has to support multiple key-type but even then I would look for alternatives (e.g. possibility to sort by key-type).

The other suggestion you give I don't fully understand, "...one account per app..." seems like a weird proposal to me given the definition of an account...

However, I do like the suggestion of selecting the key before you start connecting to another device. I personally think this is the best solution for the user experience, since it prompts the user with the question (of what key to use) when it matters. This prevents situations where the user is almost all the way through the process of exchanging keys. only to find out that they had the wrong key selected. 

Personally I would suggest to keep a single contacts list in which (if needed) it will be possible to sort by key-type.


> This resolves the problem that people who use multiple public keys, at for example different computers, will also need 2 phones and two different apps to keep their key on.

I don't agree with this reasons. If you have multiple computers you can use the same key on all of them (however, our app would have to support the possibility to provide your own key pair, but that is a different issue).

[...]

PR #59, opened 2017-05-05
Account part of the database

clenrock:
- Relevant Issues: #27 
- Related Pull Requests: #57 

* * *

## What
It adds operations to add, delete, update and retrieve the accounts list from the database.

## Why
We need a part of the database which saves account information.

## How
To test this, it's possible to run the DatabaseTest.

## Notes
IMPORTANT: this branch merged the branch for pull request #57 in. So that PR needs to be merged first, before this one can be merged.
Also, the database now uses an array of Accounts, so that in the future we can have multiple accounts/profiles/keys :wink:


ericcornelissen:
Just a quick first comment: I still feel like we need to go with a different name then "Accounts", as discussed in #50

app/src/test/java/com/nervousfish/nervousfish/modules/database/DatabaseTest.java
```diff
@@ -0,0 +1,211 @@
...
+    }
+
+    @Test
+    public void testGetAccountEmpty() throws IOException {
+        List<Account> actual = database.getAccounts();
+        assertEquals(new ArrayList<Account>(), actual);
+    }
+
+    @Test
+    public void testAddAccount() throws IOException {
```
TheBonheurs: I really like how you test account adding/deleting here 👍 . I think this should also be done with contacts in #57

clenrock:
@ericcornelissen I agree, but are we going to change that now, or the next sprint? (where we will discuss this in person)

ericcornelissen:
> I agree, but are we going to change that now, or the next sprint? (where we will discuss this in person)

I'm not going to accept this PR before we figured it out, I personally don't care too much whether that is this or next sprint 😉

clenrock:
@ericcornelissen What do you suggest as a fix?

app/src/test/java/com/nervousfish/nervousfish/modules/database/DatabaseTest.java
ericcornelissen: I'm not sure if I agree. I think that, in unittest, you want to accomplish that only one part of the functionality is tested so you can easily trace back problems. What you're doing in this test is testing both the functionality of `addAccount` and `getAccount`...

@jverbraeken, what is your opinion in this?

ericcornelissen:
> @ericornelissen What do you suggest as a fix?

I've stated my opinion in issue #50. I personally prefer the term "Profile" but I'm open to other suggestions 🙂

clenrock:
@ericcornelissen I don't see much difference in the name Account or Profile, I think they mean the same thing.

app/src/main/java/com/nervousfish/nervousfish/modules/database/GsonDatabaseAdapter.java
```diff
@@ -32,14 +35,17 @@
 /**
  * An adapter to the GSON database library
  */
+@SuppressWarnings("PMD.TooManyMethods")
 public final class GsonDatabaseAdapter implements IDatabase {
 
     private final static Type TYPE_CONTACT_LIST = new TypeToken<ArrayList<Contact>>(){}.getType();
-
+    private final static Type TYPE_PROFILE_INFORMATION = new TypeToken<ArrayList<Profile>>(){}.getType();
```
ericcornelissen: @clenrock, all caps :wink:

PR #76, opened 2017-05-09
Fix hardcoding of sizes in login and main activity

clenrock:
## What
All the hardcoded sizes are now put in the dimens.xml.

## Why
Some of the sizes are used more than once, and it is easy to have this in one place so that you don't have to change a lot of files  when you want to change the size.

## How
You can check in the code what is changed, and start the app to check that everything still looks right.


app/src/main/res/layout/content_main.xml
```diff
@@ -12,10 +12,9 @@
```
ericcornelissen: Could you look at line 10 and 11 in this file 😉 I think you can just remove them...

clenrock: Good point! Thanks

PR #78, opened 2017-05-09
Update the Coveralls integration

ericcornelissen:
- Relevant Issues: -
- Related Pull Requests: #54 

* * *

## What
This PR adds two things:
- Proper support for Coveralls, unfortunately the previous PR for coveralls didn't quite work out.  Older builds would not [do much](https://travis-ci.org/ericcornelissen/NervousFish/builds/229976329#L2865), but [now](https://travis-ci.org/ericcornelissen/NervousFish/builds/230237478#L3111) the report is uploaded to Coveralls.
- Switch from Jacoco to Cobertura.

## Why
The Coveralls support has been updated so that it actually works. The switch from Jacoco to Cobertura is to allow for this update, as I couldn't get everything working with the Jacoco report.

## How
Updated the coverage report generator (Jacoco -> Cobertura) and upgrade the Coveralls dependency.

## Alternative implementation
n/a

## Notes
You can see it working in the PR checks 👍


build.gradle
```diff
@@ -6,7 +6,8 @@ buildscript {
     }
     dependencies {
         classpath 'com.android.tools.build:gradle:2.3.1'
-        classpath 'org.kt3k.gradle.plugin:coveralls-gradle-plugin:0.3.0'
+        classpath 'org.kt3k.gradle.plugin:coveralls-gradle-plugin:2.0.0'
```
jverbraeken: Can you update this plugin to the most recent version 2.8.1?

Issue #67, opened 2017-05-08
Refractor serialization key to key class


jverbraeken:
- Parent issues: -
- Relevant issues: -
- Estimated effort: 1

* * *

## User story
As a developer I want to have the logic for serializing a key in the Key class itself so that it's less coupled to the Database class.

## What
This issue demands to have the logic for serializing a key in the Key class itself

## Why
This issue is needed for the project because we want to reduce the coupling of the Key to the Database class

## How
This feature can be added by moving the relevant code from the Database class to the Key class

## Definition of done
This task can be considered done when

- The relevant code is in the Key class

PR #81, opened 2017-05-09
Update key to JSON serialization

ericcornelissen:
- Relevant Issues: #67
- Related Pull Requests: -

* * *

## What
This PR changes two main things about the application. First and foremost it updates the way keys (instances of the `IKey` class) are serialized to JSON. As specified in #67 this is now done in the key class itself. Similarly the key class now also knows how to construct a key from given a `JsonReader`.

Besides that the JSON format for keys changed slightly to better accommodate the new writing style. To avoid having the rewrite (and possibly forget) the key type, this is still done in the `GsonKeyAdapter` class. This is now the first entry in an array representing the key. The second element is an object of key:value pairs constructed by the key class. So `{"_type":"simple","key":"key"}` changed to `["simple",{"key":"key"}]`.

## Why
This was done because the old implementation required a different class then the target class to know how to do something the target class should know itself.

## How
This was done by updating the `IKey` interface and moving the serialization code from the `GsonKeyAdapter` class to the classes implementing `IKey`.

## Alternative implementation
n/a

## Notes
I'm not quite sure about reading the keys yet (this is currently implemented using `static` methods on the classes implementing `IKey`), if anyone has a better suggestion be sure to let me know!

### DIFF
--- a/app/src/main/java/com/nervousfish/nervousfish/data_objects/IKey.java
+++ b/app/src/main/java/com/nervousfish/nervousfish/data_objects/IKey.java
@@ -1,6 +1,5 @@
 package com.nervousfish.nervousfish.data_objects;
 
-import com.google.gson.stream.JsonReader;
 import com.google.gson.stream.JsonWriter;
 
 import java.io.IOException;
--- a/app/src/main/java/com/nervousfish/nervousfish/data_objects/IKey.java
+++ b/app/src/main/java/com/nervousfish/nervousfish/data_objects/IKey.java
@@ -27,6 +27,6 @@ public interface IKey {
     /**
      * Write the key to a {@link JsonWriter}.
      */
-    void writeJSON(final JsonWriter writer) throws IOException;
+    void toJSON(final JsonWriter writer) throws IOException;
 
 }
--- a/app/src/main/java/com/nervousfish/nervousfish/data_objects/IKey.java
+++ b/app/src/main/java/com/nervousfish/nervousfish/data_objects/IKey.java
@@ -1,5 +1,9 @@
 package com.nervousfish.nervousfish.data_objects;
 
+import com.google.gson.stream.JsonWriter;
+
+import java.io.IOException;
+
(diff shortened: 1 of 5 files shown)


jverbraeken:
@ericcornelissen I like your implementation and have a suggestion to improve on your last note: https://github.com/ericcornelissen/NervousFish/commit/bda3ad09c8e35ee9038509999383e0bed353a54b
The method fromJson from RSAKey and SimpleKey is a bit dangerous because every class can call it, possibly with incorrect data (simplekey data in the fromJson of RSAKey) resulting in a runtime error. The approach in this rought draft improves on that, although it's still no ideal solution. The constructor is still public, something you generally don't want, but in this case it must be public because of restrictions of the gson library.

ericcornelissen:
@jverbraeken, how does your solution improve upon the issue that the `JsonReader` could contain incorrect JSON for the class on which it is called? It is still possible to get a `KeyCreator` from the RSA class and input a `JsonReader` which holds a `SimpleKey`, right?

app/src/main/java/com/nervousfish/nervousfish/data_objects/IKey.java
```diff
@@ -15,8 +19,13 @@
     /**
      * Get a string representation of the key type.
      *
-     * @return The type.
+     * @return The key type.
      */
     String getType();
 
+    /**
+     * Write the key to a {@link JsonWriter}.
```
clenrock: @param comment? ;)

jverbraeken:
@ericcornelissen Actually, no.
The fromJson method is package-private, so only the GsonKeyAdapter can access it and a random class from the World cannot.

app/src/main/java/com/nervousfish/nervousfish/data_objects/IKey.java
ericcornelissen: Good catch, too late though...

PR #83, opened 2017-05-09
Add tests for the data_objects module

ericcornelissen:
- Relevant Issues: -
- Related Pull Requests: -

* * *

## What
This PR adds unittests for the `data_objects` module, up to 100% coverage 🎉 Also changed some JavaDoc and fixed some issues with the `Profile` and `KeyPair` classes (`public` attributes/missing methods).

## Why
Because we need test coverage, and we forgot to test this module properly.

## How
Unittest using junit

## Alternative implementation
n/a

## Notes
n/a


app/src/test/java/com/nervousfish/nervousfish/data_objects/RSAKeyTest.java
```diff
@@ -0,0 +1,78 @@
...
+    @Test
+    public void testGetKeyReturnsNonEmptyString() {
+        IKey key = new RSAKey("foo", "bar");
+        assertNotEquals("", key.getKey());
+    }
+
+    @Test
+    public void testGetTypeReturnsRSA() {
+        IKey key = new RSAKey("foo", "bar");
+        assertEquals(ConstantKeywords.RSA_KEY, key.getType());
```
clenrock: Do you think it does not look nice to have more than 1 assert in a test? I think that is better understandable than having duplicate code for each new assert.

ericcornelissen: I'm not quite sure what you mean, but each test case should test only one thing. And the only thing I'm testing here is whether or not the `.getType()` method returns the expected value...

Issue #84, opened 2017-05-09
Database not working due to READ_ONLY access on devices/emulator


ericcornelissen:
[//]: # "Check if the title is descriptive!"
- Parent issues: -
- Relevant issues: -
- Estimated effort: 2 hours

* * *

## What
Currently the database module can't write on an actual device/emulator due to the fact that it doesn't use a dedicated Android writer/path to write to the JSON files (READ_ONLY access). For more details read [this](https://developer.android.com/training/basics/data-storage/files.html) link. Actual error from the stacktrace when starting the application on an emulator:

```
java.io.FileNotFoundException: contacts.json (Read-only file system)
  at java.io.FileOutputStream.open(Native Method)
  at java.io.FileOutputStream.<init>(FileOutputStream.java:221)
  at java.io.FileOutputStream.<init>(FileOutputStream.java:108)
  at com.nervousfish.nervousfish.modules.database.GsonDatabaseAdapter.initializeContacts(GsonDatabaseAdapter.java:298)
  at com.nervousfish.nervousfish.modules.database.GsonDatabaseAdapter.initializeDatabase(GsonDatabaseAdapter.java:286)
  ...
```

## Why
n/a

## How
t.b.d.

## Definition of done
This issue is done as soon as the database module can write to an emulator/device.

## Notes
n/a

PR #86, opened 2017-05-09
Database modification to fix Read Only error

clenrock:
Related issue: #84 

## What
It is a modification in the GsonDatabase and the Constants to fix an Read Only error we've been getting.

## Why
With the current database we can't write to files in the app itself, we are trying to write too root now.

## How
You can try to print out the database itself and have no problems, where before you got an IOException.

## Notes
The code as it is right now is probably not the nicest way to fix the problem. Maybe you think that having a non-static 'constant' in Constants is wrong, which is of course a bit weird. But in the current form with the private variables it doesn't really matter if they are static or not...
I'm open to suggestions, and would love to receive so as I already see that it is not a nice solution.


app/src/main/java/com/nervousfish/nervousfish/service_locator/EntryActivity.java
```diff
@@ -21,7 +21,8 @@
     protected void onCreate(final Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
 
-        final IServiceLocator serviceLocator = ServiceLocatorCreator.createInstance();
+        final IServiceLocator serviceLocator = ServiceLocatorCreator.createInstance(
```
ericcornelissen: It seems odd to me that this works, but I think I'll be able to come up with a solution that is a bit nicer given this possibility 😃 Is it OK if I continue developing on this branch?

clenrock: Yes, that is fine!

ericcornelissen: After some reconsideration, I think it is better to do this on a separate branch so we can start using the database in the meantime 👍

PR #90, opened 2017-05-10
Refractoring the ServiceLocator, closes #60

### LINKED ISSUE #60: Omit the use of Greenrobot.eventbus to distribute the ServiceLocator
- Parent issues: -
- Relevant issues: -
- Estimated effort: 3

* * *

## What
Remove the `ServiceLocator` + `ServiceLocatorCreater` construct where the classes that the `ServiceLocator` instantiates need to be told about the existence of the `ServiceLocator` via a communication bus.

## Why
Because it is incredibly cumbersome. Each of the classes mentioned above now requires three methods just for instantiation, and proper instantiation becomes very hard because the dependencies the `ServiceLocator` provides cannot be set to `final` attributes (since they can't be accessed in the constructor).

## How
My proposal is to remove the `ServiceLocatorCreator` entirely, make the `ServiceLocator` itself a singleton-like class and provide `this` when the `ServiceLocator` instantiates other classes.

## Definition of done
This is done once there is no longer any need for the greenrobot eventbus to distribute the ServiceLocator AND the ServiceLocator is provided via the constructor of relevant classes.

## Notes
n/a

jverbraeken:
- Relevant Issues: #60 
- Related Pull Requests: ...

* * *

## What
This PR removes the ServiceLocatorCreator and gives modules the opportunity to access other modules over the service locator directly.

## Why
This is needed because we accessed unitialized modules in the constructor last week.

## How
If the app still runs, then it works fine ;-)

## Alternative implementation
I talked with Eric about many alternatives, and we agreed that this solutions would be the best.

## Notes
-


app/src/main/java/com/nervousfish/nervousfish/modules/database/GsonDatabaseAdapter.java
```diff
@@ -40,47 +34,27 @@
 @SuppressWarnings("PMD.TooManyMethods")
 public final class GsonDatabaseAdapter implements IDatabase {
 
-    private final static Type TYPE_CONTACT_LIST = new TypeToken<ArrayList<Contact>>(){}.getType();
-    private final static Type TYPE_PROFILE_LIST = new TypeToken<ArrayList<Profile>>(){}.getType();
+    private final static Type TYPE_CONTACT_LIST = new TypeToken<ArrayList<Contact>>() {
+    }.getType();
+    private final static Type TYPE_PROFILE_LIST = new TypeToken<ArrayList<Profile>>() {
+    }.getType();
```
ericcornelissen: Could you format these a little better 😇

app/src/main/java/com/nervousfish/nervousfish/modules/database/GsonDatabaseAdapter.java
```diff
@@ -40,47 +34,27 @@
...
     private final static String CONTACT_NOT_FOUND = "Contact not found in database";
-    @SuppressWarnings("PMD.SingularField")
-    private final IServiceLocatorCreator serviceLocatorCreator;
-    private String contactsPath;
-    private String profilesPath;
+    @SuppressWarnings("PMD.SingularField") // This constants will be used later on probably
+    private final IConstants constants;
```
ericcornelissen: Just replace it with a local `final` variable in the constructor

app/src/main/java/com/nervousfish/nervousfish/service_locator/IServiceLocator.java
```diff
@@ -25,4 +25,11 @@
     INFCHandler getNFCHandler();
     IQRHandler getQRHandler();
 
+    /**
+     * Registers the class specified to the EventBus
+     *
+     * @param object The Object that should be registed to the EventBus
+     */
+    void registerToEventBus(final Object object);
```
ericcornelissen: Why is this needed, it is currently unused?

app/src/main/java/com/nervousfish/nervousfish/service_locator/ServiceLocator.java
```diff
@@ -24,89 +35,108 @@
...
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void registerToEventBus(final Object object) {
+        EventBus.getDefault().register(object);
+    }
+
+    private void checkModule(final IModule module, final String moduleName) {
```
ericcornelissen: Missing JavaDoc

app/src/main/java/com/nervousfish/nervousfish/modules/database/GsonDatabaseAdapter.java
jverbraeken: I'm sorry Eric, but no... I cannot I think :-8
If I change it back and do auto format in Intelllij, then it's changed back to this.

app/src/main/java/com/nervousfish/nervousfish/service_locator/IServiceLocator.java
jverbraeken: Within a few days it will be used, for example to let the BluetoothActivity know that there were new Bluetooth devices detected or to let the NFC Activity know that some device touched your OnePlus 3

ericcornelissen: 👍

app/src/main/java/com/nervousfish/nervousfish/modules/database/GsonDatabaseAdapter.java
ericcornelissen: imo that shouldn't be a reason but fine 😕 Could you at least do it as follows to make it a little bit less ugly (if the execution line continues, it should start at a higher indentation)
```java
private final static Type TYPE_CONTACT_LIST =
            new TypeToken<ArrayList<Contact>>() {
            }.getType();
private final static Type TYPE_PROFILE_LIST =
            new TypeToken<ArrayList<Profile>>() {
            }.getType();
```

Issue #68, opened 2017-05-08
Add ContactActivity containing contact information


jverbraeken:
- Parent issues: -
- Relevant issues: -
- Estimated effort: 8

* * *

## User story
As a user I want to see all information I have about a certain contact, including his public keys, when I click on a contact on the contact list

## What
This issue demands an overview of all data the user has about a specific contact that can be reached by clicking on a contact

## Why
This issue is needed for the project because the user may want to view some data of a contact, like the telephone number or e-mail address

## How
This feature can be added by adding a new ContactActivity containing all Widgets needed to give the user a complete overview of all data he/she has about a specific user

## Definition of done
This task can be considered done when

- The user can access the data of a specific contact by clicking on a contact
- All relevant data of the contact is displayed in an intuitive way
- The data displayed corresponds to the database
- The data can be changed by the user
- The changes are reflected in the database

PR #93, opened 2017-05-10
Allow contacts to have multiple keys, connects to #68

ericcornelissen:
- Relevant Issues: #68
- Related Pull Requests: -

* * *

## What
This PR adds functionality to allow Contacts to have multiple keys.

## Why
Because your friends might have multiple keys and it can be useful to have them grouped together.

## How
There are new test suites verifying its behavior, furthermore the functionality is not yet used and can thus not be seen.

## Alternative implementation
n/a

## Notes
n/a


ericcornelissen:
I updated the test and they're now working, so I guess it does 😄 If you have you're doubts you can run and review the tests locally to see if you agree

app/src/test/java/com/nervousfish/nervousfish/data_objects/ContactTest.java
```diff
@@ -34,9 +47,16 @@ public void testGetNameReturnsTheName() {
     }
 
     @Test
```
TheBonheurs: Maybe add a test with multiple keys returned?

PR #91, opened 2017-05-10
Change the ListView on the MainActivity, closes #10

clenrock:
- Relevant Issues: #10 

* * *

## What
A nicer contacts list with their own Adapter, so changing it will be easier in the future. This will also look in the database for the contacts instead of getting it from an ArrayList with dummy data.

## Why
It gives a nicer user experience and it had to be connected to the database.

## How
When you run the app and go to the MainActivity you can see the new list.

## Notes
In the future we should try to find a solution for the grey part below the list, when you don't have a lot of contacts yet.

PR #95, opened 2017-05-10
Create a back and delete button on the ContactActivity, connects to #68

clenrock:
- Relevant Issues: #68 
- Related Pull Requests: -

* * *

## What
This adds two buttons to the ContactActivity. One which will get you back to the previous Activity, and one to delete the Contact you are currently on.

## Why
We needed a way to get back to the previous Activity, and to delete a Contact.

## How
When you open the app, press on one of the list items in the MainActivity to go to the ContactPage. Here you can press the buttons on the top.

## Notes
The delete button doesn't delete the contact yet, when #91 and the part of showing the information of a certain contact is on develop, I will fix this.


app/build.gradle
```diff
@@ -42,6 +42,10 @@ android {
     lintOptions {
         disable 'GoogleAppIndexingWarning'
     }
+
+    repositories {
+        mavenCentral()
```
ericcornelissen: The repositories to use should be specified in the `build.gradle` in the root of the repo ([link](https://github.com/ericcornelissen/NervousFish/blob/develop/build.gradle#L15))

app/build.gradle
```diff
@@ -51,6 +55,7 @@ dependencies {
     compile 'com.android.support:design:25.3.1'
     compile 'com.google.code.gson:gson:2.7'
     compile 'org.greenrobot:eventbus:3.0.0'
+    compile 'cn.pedant.sweetalert:library:1.3'
```
ericcornelissen: Have you consider using the Android build-in [Dialogs](https://developer.android.com/guide/topics/ui/dialogs.html)/[PopupWindows](https://developer.android.com/reference/android/widget/PopupWindow.html). I'm in favor of using these over some custom library for consistency with other Android apps...

clenrock: Yes, and I've even build a popup myself. But then I encountered this library, which makes it much easier and cleaner to create popups

app/build.gradle
clenrock: All right!

app/src/main/java/com/nervousfish/nervousfish/activities/ContactActivity.java
```diff
@@ -29,6 +33,33 @@ protected void onCreate(final Bundle savedInstanceState) {
...
+            public void onClick(final View v) {
+                finish();
+            }
+        });
+
+        final ImageButton deleteButton = (ImageButton) findViewById(R.id.deleteButton);
+        deleteButton.setOnClickListener(new View.OnClickListener() {
+            public void onClick(final View v) {
+                new SweetAlertDialog(ContactActivity.this, SweetAlertDialog.WARNING_TYPE)
+                        .setTitleText("Are you sure?")
```
jverbraeken: Can you put these strings in strings.xml instead of hardcoding it? If I remember correctly you can use getResources().getString(R.string.foobar)

app/build.gradle
```diff
@@ -3,6 +3,10 @@ apply plugin: 'cobertura'
 apply plugin: 'com.github.kt3k.coveralls'
 apply from: '../config/quality.gradle'
 
+repositories {
+    mavenCentral()
```
jverbraeken: You don't need these 3 lines

ericcornelissen:
Could you also add at least one Cucumber test for the additions in this PR? 😇

clenrock:
@ericcornelissen Good point!

app/src/main/java/com/nervousfish/nervousfish/activities/ContactActivity.java
clenrock: Thanks! I will do that

app/build.gradle
clenrock: All right, I'll try to remove them

Issue #74, opened 2017-05-08
Expand the database with more keys for a contact


clenrock:
- Parent issues: #27 
- Relevant issues: #68 
- Estimated effort: 2 (Only Cornel's estimation)

* * *

## User story
As a user I want to be able to save more than 1 public key per contact so that if the other person has more than 1 key I can save all of them

## What
An expanded database which has an array of public keys instead of only 1.

## Why
A lot of people have more than 1 public keys to send them different kinds of data encrypted, so it would be easy to have them all under 1 contact in our app.

## How
In our JSON database there is a field in the contact with the public key, this should be changed to an array of public keys. For this a couple of methods in the GSONDatabaseAdapter should be changed.

## Definition of done
When it is possible to save more than 1 public key per contact.

Issue #69, opened 2017-05-08
Use user's contact in app


jverbraeken:
- Parent issues: -
- Relevant issues: -
- Estimated effort: -

* * *

## User story
As a user I want that the app uses the contacts I already have so that I can easily see who of my contacts has his/her public key shared with me

## What
This issue demands to use the contacts that already on the phone, synchronized by Google, as the contacts in the app

## Why
This issue is needed for the project because the user may want to pair the public key with a contact he/she already has so that he/she has direct access to all other data he has about the person he pairs with.

## How
This feature can be added by synchronizing the users contacts with the application

## Definition of done
This task can be considered done when

- The user's contact are drawn at the contact screen
- The app makes clear of whose contacts the user already has a public key and from whose contacts he/she doesn't

PR #99, opened 2017-05-12
Give keys and keypairs a name

ericcornelissen:
- Relevant Issues: #69, #74
- Related Pull Requests: #89, #93

* * *

## What
Allow both the `IKey` class and `KeyPair` class to have names so they're easier to identify.

## Why
This allows users the identify their keys by name instead of the key itself.

## How
Adding a `private final String` for the name to both the classes and requiring them to be set in the constructor.

## Alternative implementation
Non

## Notes
n/a


app/src/main/java/com/nervousfish/nervousfish/data_objects/KeyPair.java
```diff
@@ -14,12 +15,22 @@
      * @param publicKey  The publicKey of the KeyPair.
      * @param privateKey The privateKey of the KeyPair.
      */
-    public KeyPair(final IKey publicKey, final IKey privateKey) {
+    public KeyPair(final String name, final IKey publicKey, final IKey privateKey) {
+        this.name = name;
```
jverbraeken: Add javadoc

app/src/main/java/com/nervousfish/nervousfish/data_objects/KeyPair.java
```diff
@@ -56,7 +68,7 @@ public boolean equals(final Object o) {
      */
     @Override
     public int hashCode() {
-        return this.publicKey.hashCode() + this.privateKey.hashCode();
+        return this.name.hashCode() + this.publicKey.hashCode() + this.privateKey.hashCode();
```
jverbraeken: We should consider adding HashCodeBuilder and EqualsBuilder for next week

app/src/main/java/com/nervousfish/nervousfish/data_objects/KeyPair.java
ericcornelissen: Good catch!

app/src/main/java/com/nervousfish/nervousfish/data_objects/KeyPair.java
TheBonheurs: I'm mot sure this is needed as that is very class specific

PR #73, opened 2017-05-08
Add the Contact Activity, opened after clicking a contact

clenrock:
- Relevant Issues: #68 

* * *

## What
This pull request adds an activity which is the start of a contact page. 

## Why
The contact page should have the info about a certain person and have a list with his public keys.

## How
It can be tested by changing the 'intent' in EntryActivity to ContactActivity. This way you can see the contact page.

## Notes
This pull request is only for the activity, it retrieves nothing from the database and is not yet accessible from the MainActivity. So mostly look at the design and note things that should look nicer or can be added.
image

PR #100, opened 2017-05-12
Open contact details in contact activity

ericcornelissen:
- Relevant Issues: #68 
- Related Pull Requests: #73, #93, #99

* * *

## What
With this PR it is possible to click on a contact in the main activity and view its details in the contact activity.

## Why
Because we don't want to see some example data on the contact activity, but we want to actually see the data of the selected user.

## How
The selected contact is given from the main activity to the contact activity via an intent.

To see this you can simply open the app, click a contact and see if it works.

## Alternative implementation
Non.

## Notes
n/a


app/src/main/java/com/nervousfish/nervousfish/activities/ContactActivity.java
```diff
@@ -19,19 +24,20 @@
...
      * Creates the new activity, should only be called by Android
      *
-     * @param savedInstanceState Don't touch this
+     * @param savedInstanceState The saved state of the instance.
      */
     @Override
     protected void onCreate(final Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
-        setContentView(R.layout.activity_contact);
+        this.setContentView(R.layout.activity_contact);
```
jverbraeken: Officially it is super.setContentView, but it's ok

app/src/main/java/com/nervousfish/nervousfish/activities/MainActivity.java
```diff
@@ -46,77 +51,94 @@
     @Override
     protected void onCreate(final Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
-        setContentView(R.layout.activity_main);
-        LOGGER.info("MainActivity created");
```
TheBonheurs: No logger?

app/src/main/java/com/nervousfish/nervousfish/activities/ContactActivity.java
```diff
@@ -59,22 +65,32 @@ public void onClick(final SweetAlertDialog sDialog) {
...
-
-        lv.setAdapter(new ArrayAdapter<>(this,
-                android.R.layout.simple_list_item_1, keys));
-
+    /**
+     * Set the name of the {@link Contact} to the {@link ContactActivity}.
+     *
+     * @param contact The {@link Contact}.
+     */
+    private void setName(final Contact contact) {
```
clenrock: I don't find this really intuitive, I would just make the parameter a string, or do you think otherwise?

app/src/main/res/layout/contact_list_entry.xml
```diff
@@ -24,15 +24,9 @@
         <TextView
             android:id="@+id/name"
             android:layout_width="match_parent"
-            android:layout_height="wrap_content"
+            android:layout_height="match_parent"
             android:text="@string/name"
-            android:textSize="@dimen/contacts_list_entry_text_name_size" />
-
-        <TextView
-            android:id="@+id/pubKeySnippet"
```
clenrock: I still think we should keep this... Maybe you know the first couple of characters of someone's key, and this gives more confidence you got the right person

app/src/main/java/com/nervousfish/nervousfish/activities/MainActivity.java
ericcornelissen: Moved it to the end of the method, see [line 100](https://github.com/ericcornelissen/NervousFish/pull/100/files#diff-149538cc1f1f5e4370b6ae93d6489330R99)

app/src/main/java/com/nervousfish/nervousfish/activities/ContactActivity.java
ericcornelissen: I agree 😅 missed that, it originally just set the name of the contact...

app/src/main/res/layout/contact_list_entry.xml
ericcornelissen: I disagree because of three reasons:
- A contact can have multiple keys, in that case which key should you display?
- I don't think anyone really knows the public key of their friend by heard, so it won't give you any confidence you're selecting the right person.
- There is really no need for the user to see the key except for very specific/rare cases. In these cases it should not be an issue two have to perform two additional clicks (on the contact -> on the key of interest in the `ContactActivity`).

app/src/main/java/com/nervousfish/nervousfish/activities/ContactActivity.java
ericcornelissen: I can agree with that, I'll change it 👍

app/src/main/java/com/nervousfish/nervousfish/activities/MainActivity.java
```diff
@@ -117,6 +142,7 @@ private void fillDatabaseWithDemoData() {
             e.printStackTrace();
```
TheBonheurs: Log error maybe?

ericcornelissen: That is a good one. But since it is not done anywhere yet I'll create a separate PR for that where I add a log for all instances of `.printStackTrace();` 👍

ericcornelissen: Opened a new PR for this, see #108 :wink:

PR #113, opened 2017-05-15
Improve implementation of keys to future proof code

ericcornelissen:
- Relevant Issues: -
- Related Pull Requests: -

* * *

## What
Replaced the static methods to create a new `RSAKey` or `SimpleKey` given a `JsonReader` with a constructor that accepts a `Map` that should contain the relevant keys.

Also added two new static methods to the `RSAKey` class which can be used to get the `exponent` and `modulus` of an `RSAKey`.

## Why
The constructor change is a more intuitive solution for the problem and the `RSAKey` static methods for the exponent/modulus might become useful in the future for distributing keys to other apps.

## How
You can run the unittests to verify that all the code in `data_objects` is still covered so it should work correctly

## Alternative implementation
None

## Notes
None

### DIFF
--- a/app/src/main/java/com/nervousfish/nervousfish/data_objects/RSAKey.java
+++ b/app/src/main/java/com/nervousfish/nervousfish/data_objects/RSAKey.java
@@ -38,10 +38,14 @@ public final class RSAKey implements IKey {
      *
      * @param map A {@link Map} mapping {@link RSAKey} attribute names to values.
      */
-    public RSAKey(final Map<String, String> map) {
+    public RSAKey(final Map<String, String> map) throws IllegalArgumentException {
         this.name = map.get(RSAKey.JSON_CONSTANT_NAME);
         this.modulus = map.get(RSAKey.JSON_CONSTANT_MODULUS);
         this.exponent = map.get(RSAKey.JSON_CONSTANT_EXPONENT);
+
+        if (this.name == null || this.modulus == null || this.exponent == null) {
+           throw new IllegalArgumentException();
+        }
     }
 
     /**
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/nervousfish/nervousfish/data_objects/RSAKey.java
```diff
@@ -36,24 +34,46 @@ public RSAKey(final String name, final String modulus, final String exponent) {
...
-     * @throws IOException When the {@link JsonReader} throws an {@link IOException}.
+     * @param map A {@link Map} mapping {@link RSAKey} attribute names to values.
      */
-    static public IKey fromJSON(final JsonReader reader) throws IOException {
-        final Map<String, String> map = new ConcurrentHashMap<>();
-        while (reader.hasNext()) {
-            final String name = reader.nextName();
-            final String value = reader.nextString();
-            map.put(name, value);
+    public RSAKey(final Map<String, String> map) {
```
jverbraeken: Can you check if the map contains these strings and otherwise log it or even throw an exception?

app/src/main/java/com/nervousfish/nervousfish/data_objects/SimpleKey.java
```diff
@@ -29,23 +27,13 @@ public SimpleKey(final String name, final String key) {
...
-            final String name = reader.nextName();
-            final String value = reader.nextString();
-            map.put(name, value);
-        }
-
-        final String name = map.get("name");
-        final String key = map.get("key");
-        return new SimpleKey(name, key);
+    public SimpleKey(final Map<String, String> map) {
+        this.name = map.get("name");
```
jverbraeken: same

PR #114, opened 2017-05-15
Pairing button, closes #52

### LINKED ISSUE #52: The + button shows a menu with Bluetooth, NFC and QR
[//]: # "Check if the title is descriptive!"
- Parent issues: -
- Relevant issues: -
- Estimated effort: 3

* * *

## User story
As a user I want the + button to show new buttons to add a new public key over Bluetooth, NFC and QR and I want the + button to hide the new buttons when it's clicked while the new buttons are displayed.

## What
An intuitive, user-friendly and clean way to make it possible to add new public keys using various connection mechanisms

## Why
This makes the interface look like a lot "cleaner" than when all buttons are visible initially

## How
To be found out...

## Definition of done
When the + button is clicked 3 new buttons appear above the + button. One contains a bluetooth logo, one contains an NFC logo and one contains a camera logo. When the + button is clicked again the new buttons disappear.

## Notes
-

jverbraeken:
- Relevant Issues: #52 
- Related Pull Requests: ...

* * *

## What
This PR adds a button to the right bottom corner of the MainActivity. When clicked, a menu opens with 3 buttons: bluetooth, nfc and qr.

## Why
This PR is needed for this project because the user must be able to choose between the various pairing methods when we offer these.

## How
This PR can be tested by clicking on the pairing button.

## Alternative implementation
I checked out quite some libraries and I chose Clang's library because it was quite flexible and easy-to-use.

## Notes
-


ericcornelissen:
Why did you change the layout of the list view in the main activity? Could you at least change the background two white (or the background of the images to gray)

Also, why waste the main button itself for closing the menu? The menu will close automatically when you click outside the menu... and even if this is not intuitive I think it will not be very common that someone accidentally clicks the menu button in which case clicking the back button (which IS intuitive) will do. I would change the button to open the menu to a usable button (e.g. the Bluetooth button) when the menu is opened.

jverbraeken:
@ericcornelissen Thanks for your review! For some reason I had to run the app 3 times (?!?) before I saw what you meant with the layout of the MainActivity, was already typing an answer back that you probably made some mistake, but, anyway, something went wrong on my machine apparently.

I understand your second point and I agree with you that what you propose is nicer, but I couldn't find a library that could do that. I looked at

- https://github.com/wangjiegulu/RapidFloatingActionButton
- https://github.com/Clans/FloatingActionButton
- https://github.com/makovkastar/FloatingActionButton
- https://github.com/futuresimple/android-floating-action-button
- https://github.com/pregiotek/ion-floating-menu

And apart from that I think it's much more important to put my time in functionality than design right now.

ericcornelissen:
I would say you can just change the `onClick` event when the button is being pressed, but I might be thinking too simple there...

Anyway, I agree that you might have better things to do and this visual thing can be resolved later 😄

app/src/main/res/values/strings.xml
```diff
@@ -8,6 +8,9 @@
     <string name="title_activity_contact">Contact</string>
     <string name="msg_mismatch">Your password is wrong</string>
     <string name="contact_title">Contact</string>
+    <string name="qr">Camera</string>
```
ericcornelissen: Why not just call this "QR"?

app/src/main/res/anim/jump_from_down.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<set xmlns:android="http://schemas.android.com/apk/res/android"
+    android:interpolator="@android:interpolator/overshoot">
+    <alpha
+        android:fromAlpha="0"
+        android:toAlpha="1"
+        android:duration="400" />
```
TheBonheurs: Why 4 seconds to jump up but 2 seconds to jump down?

app/src/main/res/values/strings.xml
```diff
@@ -8,6 +8,9 @@
...
+    <string name="nfc">NFC</string>
+    <string name="bluetooth">Bluetooth</string>
 
     <string name="name">Name</string>
     <string name="public_key_snippet">Public Key Snippet</string>
```
TheBonheurs: Here in line 21 place a space after the comma 👺

app/src/main/res/values/strings.xml
TheBonheurs: Agree

jverbraeken: done

app/src/main/res/values/strings.xml
jverbraeken: Fixed!

TheBonheurs: 👌

Issue #107, opened 2017-05-12
Identity check with grid


jverbraeken:
- Parent issues: #104
- Relevant issues: #109
- Estimated effort: 20

* * *

## User story
As a user I want to be sure that I am connected to the intended person and not to some Man In The Middle, so that I can securely exchange public keys.

## What
This issue demands that we create a method for pairing with Bluetooth that makes sure you are connected to a certain person. This makes the connection more secure and you can be sure that there is no Man In The Middle, who can't be trusted.

## Why
This issue is needed for the project because it's really important to be able to verify that you're really paired with the correct person. Otherwise a MITM (see #80) could be possible.

## How
This feature can be added by letting the user with whom the initiating is pairing click on specific positions on a grid. For example, if a house is drawn over the grid, the user should e.g. first tap on the window, then on the door and finally on the roof.

## Definition of done
This task can be considered done when

- The exact way of how this pairing method should work is thought out and documented #144
- After a pairing attempt the target user sees a grid with an image overlay #135
- The target user can tap in a certain sequence on positions of the grid #122
- These touch events are registered #122
- These touch events are sent to the initiating user #118
- The initiating user processes these touch events and verifies whether or not these are correct. #145

PR #122, opened 2017-05-15
Add an initial visual identity check activity

ericcornelissen:
- Relevant Issues: #107
- Related Pull Requests: -

* * *

## What
A simple initial activity which can be used to verify the identity of another user while connecting via Bluetooth in a visual manner. For now the activity only contains a grid of buttons, later on the buttons will be overlayed with an image to present a more user friendly interface. Each button can be identified and after clicking on 5 buttons the activity will automatically go to the next activity (currently the LoginActivity) and provide it with the selected pattern.

## Why
As specified in #107 we need to create a secure way to pair with Bluetooth on top of the standard pairing method. This PR contains the first step towards creating this method.

## How
You can run the Cucumber tests and see it there or you can force the code to open the new activity by adding:
```java
Intent intent = new Intent(this, VisualVerificationActivity.class);
this.startActivity(intent);
```

## Alternative implementation
Not considered

## Notes
As mentioned before this PR only adds simple activity with some buttons. In future PRs there will be a few additions, for example:
- Have a 'waiting' activity for the slave while it is waiting for the master to input the pattern.
- Have a 'waiting' activity for the master while it is waiting for the slave to reproduce the pattern.
- Have an image on top of the buttons so the user can more easily specify what button it clicked (by specifying visually where they clicked).


jverbraeken:
Nice work, Eric!

Have you already looked at another much simpler solution: drawing a single image and then using [touch events](http://stackoverflow.com/questions/3476779/how-to-get-the-touch-position-in-android) to get the location of where the user pressed. Such a solution is more elegant, less code and has a lot (!) more precision than 16 buttons (each having 2560x1440 / 16 = 230,400 times less precision than OnTouchEvent(x, y))

ericcornelissen:
@jverbraeken I don't think we need that kind of precision to make this work, and doing it that way could complicate things significantly for no good reason.

The reason I don't think we need that kind of precision is because of the image that I image will be displayed on top of the buttons. This should be an image with a few distinct figures/drawings that can easily be communicated. These figures will be pretty large so it easy for user to identify them on the screen and touch them in order to replicate the pattern of the master device.

There is really only one downside to this approach, which is that all images have to adjusted for the buttons so that the distinct elements are within the borders of the underlying buttons.

app/src/main/java/com/nervousfish/nervousfish/activities/LoginActivity.java
```diff
@@ -59,7 +59,7 @@ private void validateAccount() {
         if (skipPassword) {
             LOGGER.warn("Password skipped!");
             mError.setVisibility(View.GONE);
-            final Intent intent = new Intent(LoginActivity.this, MainActivity.class);
+            final Intent intent = new Intent(LoginActivity.this, VisualVerificationActivity.class);
```
clenrock: Could you change this back? It's better to only keep that for testing

ericcornelissen: Sure, forgot about it 😅

app/src/androidTest/assets/features/visualVerification.feature
```diff
@@ -0,0 +1,18 @@
...
+  Do several things on the VisualVerificationActivity
+
+  Scenario Outline: Fill out a correct pattern
+    Given I am viewing the visual verification activity
+    When I press button <first_button>
+    And I press button <second_button>
+    And I press button <third_button>
+    And I press button <fourth_button>
+    And I press button <fifth_button>
+    Then I leave the visual verification activity
```
TheBonheurs: Should you always leave this activity, even in the case your sequence wasn't the right one? Or is this for the person that initiates the connection?

app/src/main/java/com/nervousfish/nervousfish/activities/MainActivity.java
```diff
@@ -77,7 +77,12 @@ public void onClick(final View view) {
         }
 
         final ListView lv = (ListView) findViewById(R.id.listView);
-        lv.setAdapter(new ContactListAdapter(this, this.contacts));
+        try {
+            fillDatabaseWithDemoData();
+            lv.setAdapter(new ContactListAdapter(this, serviceLocator.getDatabase().getAllContacts()));
+        } catch (final IOException e) {
+            LOGGER.error("Failed to retrieve contacts from database", e);
```
TheBonheurs: Filling the db could also go wrong,I don't mind this log but a more ambiguous one might be better.

app/src/main/java/com/nervousfish/nervousfish/activities/VisualVerificationActivity.java
```diff
@@ -0,0 +1,100 @@
...
+                    nextActivity();
+                } else {
+                    securityCode += button;
+                    LOGGER.info("code so far: " + securityCode);
+                }
+            }
+
+        };
+
+        this.findViewById(R.id.button00).setOnClickListener(onClickListener);
```
TheBonheurs: You could also use this inthe layout xml https://developer.android.com/reference/android/view/View.html#attr_android:onClick

app/src/androidTest/assets/features/visualVerification.feature
ericcornelissen: For now this is intended for the user initiating the connection

app/src/main/java/com/nervousfish/nervousfish/activities/MainActivity.java
ericcornelissen: I agree, but the `fillDatabaseWithDemoData();` should be removed in the future, thats why I exluded it from the log

TheBonheurs: Fine by me 👍

app/src/main/java/com/nervousfish/nervousfish/activities/VisualVerificationActivity.java
ericcornelissen: Good one 👍

Issue #109, opened 2017-05-12
Identity check with rhythm over Bluetooth


jverbraeken:
- Parent issues: #104
- Relevant issues: #107 #116 #117 #118 #119 #120 #121
- Estimated effort: 20

* * *

## User story
As a user I want to be sure that I am connected to the intended person and not to some Man In The Middle, so that I can securely exchange public keys.

## What
This issue demands that we create a method for pairing with Bluetooth that makes sure you are connected to a certain person. This makes the connection more secure and you can be sure that there is no Man In The Middle, who can't be trusted.

## Why
This issue is needed for the project because it's really important to be able to verify that you're really paired with the correct person.

## How
This feature can be added by letting the users that want to pair, sync their rhythm of tapping on the screen. This can be done by using, for example, a tune that the user can make or by some help of the app. How this should work should be thought out.

## Definition of ready
This features is not ready until the following issues are finished:

 - Setup a Blutooth connection #2 
## Definition of done
This task can be considered done when

- The master has an activity for tapping the rhythm #116 
- The master has an activity for waiting on the client to finish tapping his rhythm #121
- The slave has an activity for tapping the rhytm #117 
- The tapping data is sent from the slave to the master #118 
- The master verifies if the slave has tapped the same rhythm #119 
- If yes, than the new user is added to the database for both users #120 #127 
- If no, than nothing happens and the client can try it again

PR #128, opened 2017-05-15
Add an activity diagram of the rhythm pairing

jverbraeken:
- Relevant Issues: #109
- Related Pull Requests: -

* * *

## What
This PR adds an activity diagram of the rhythm pairing to the repository

## Why
This is done because the rhythm pairing #109 is pretty complex

## How
I made an activity diagram with GenMyModel

## Alternative implementation
-

## Notes
-


clenrock:
I'm not really sure about the diagram, it's confusing that you sometimes use 'master' and 'slave' but I also see 'client' being used. 
The note with 'top = master' is not fully visible, the bottom is wrapped.
Last point; why do you have the 'start app' block? I think we can assume the apps are already opened when they are pairing

ericcornelissen:
> I think we can assume the apps are already opened when they are pairing

I think we cannot just assume that, I think that must be true.

* * *

My initial response:
- I think its easier to just send the correct tab pattern to the slave and verify it on the slave machine, instead of sending the pattern of the slave back to the master.
- I think you can leave out the part after sending the public key as this is not specific to this scenario.
- I think its better that when the pattern of the slave is incorrect, the connection is stoppend and the users have to restart the connection. But I'm interested as to why you choose to take this approach.

TheBonheurs:
I also have a few questions. First of all I think pairing should be included by both, otherwise how do you know who the master and who the slave is? Next up, what does wait for the tap on the screen mean? Wait for the tap from the master or something else? After checking the rhythm and finding out it's wrong shouldn't the master re-tap a (new) rhythm?

jverbraeken:
@clenrock @ericcornelissen @TheBonheurs Thanks for your reviews!

jverbraeken:
@clenrock I changed client to slave. The note should be fully visible now, I think that the start app block is, as @ericcornelissen pointed out, needed, but we can discuss about that

jverbraeken:
@ericcornelissen I think it's dangerous to send the correct pattern to the slave, because a MITM can just very, very easy just say that his pattern matches.
I'm not sure if we should leave out the part after sending the public keys, as you can also see under the definition of done of #109. My goal was to have a fully functional public key exchange this or next week, and when you cannot add contacts, the feature is not usable at all.

My idea about tapping the rhythm is that when the user makes a mistake while tapping he can just immediately start over again without having to touch any button. Will speak about that during the meeting

jverbraeken:
@TheBonheurs I'm not sure about your first point, will speak to you about that during the meeting. Wait for a tap on the screen on the top means that the app waits for a tap from the master on the (almost) screen-filling button @clenrock made. About your last point, same as @ericcornelissen , not sure what the best way is. We'll have to discuss that

ericcornelissen:
> I think it's dangerous to send the correct pattern to the slave, because a MITM can just very, very easy just say that his pattern matches.

Fair point 😄 


> I'm not sure if we should leave out the part after sending the public keys, as you can also see under the definition of done of #109. My goal was to have a fully functional public key exchange this or next week, and when you cannot add contacts, the feature is not usable at all.

You don't have to leave it out entirely, but the details are not needed here imo since this diagram is about the rhythm pairing. You're also leaving out opening the app and selecting the pairing method...


> My idea about tapping the rhythm is that when the user makes a mistake while tapping he can just immediately start over again without having to touch any button.

This is possible, but not very intuitive imo so you have to work this out a little better.

jverbraeken:
@ericcornelissen Updated the diagram!

clenrock:
@jverbraeken  It still includes the 'start app' part

ericcornelissen:
A few comments on clearness of the text:
- `Pair over Bluetooth` should be `Select pair over Bluetooth`
- `Tab rhythm` should be `Select tab rhythm`


Also, I think the diagram should be a bit more linear... I created a quick and dirty example for you (not fully thought through ofcourse) 😄 
quick dirty

jverbraeken:
@clenrock Eric disagrees with you:

> I think we cannot just assume that, I think that must be true.

ericcornelissen:
I actually changed my mind on that. We should keep the start app balloon, but there should also be a branch where the app is already opened.

I think that whenever someone wants to connect with you you get a notification for this and clicking on this will start the pairing process.

jverbraeken:
@clenrock @ericcornelissen Implemented your suggestions!

ericcornelissen:
Still have a few remarks 😅 
- I think the slave should already be notified as soon as it is selected from the Bluetooth devices list. So it should be moved a bit to the left and there should be some kind of arrow indicating that it is being notified.
- The exchange of keys at the end of the diagram looks really weird... I understand the diagram and it seems correct, but due to the fact that "Send public key to master" and "New contact" are lined up so perfectly it kinda feels like the master only gets the slaves public key once the key accepted the contact... Maybe spread this out a bit 😄

jverbraeken:
@ericcornelissen Implemented your suggestions!

ericcornelissen:
Nice update! Got two remaining comments:
- Shouldn't the "Waiting for slave" and "Wait for taps from slave" be after one another? Are these really supposed to happen in parallel?
- Did you think about what should happen if at any point in time the connection fails? Most interesting, imo, would be the situation where the master send his details to the slave but then the connection breaks before the slave can respond with their details.
  In that case, do we want the slave to be able to add the master as a contact, I'm not so sure...? In this scenario I would image the master and slave would want to retry the exchange from the start, otherwise the master can't receive the details of the slave. In which case already having the master as contact can be annoying...

jverbraeken:
@ericcornelissen I implemented your first suggestion, but I'm not sure about the second one. The diagram will become very messy when I indicate what should happen in case of a connection failure for all nodes where data is sent over Bluetooth. I think it's best to just show the user an error message when onBluetoothConnectionLost is called and then give the user the option if he wants to retry or to cancel the public key exchange. 

If A received the public key from B, but due to a network failure B didn't receive the public key from A, then A can still add person B imo. Public keys are public, it's no problem if you know the public key of some person while that person doesn't know the public key of you.

PR #135, opened 2017-05-17
Add proper image overlay to visual verification grid

ericcornelissen:
- Relevant Issues: #107 
- Related Pull Requests: #122 

* * *

## What
Adds a proper image overlay for the grid introduced in #122. Also reduced the number of buttons in the grid from 16 to 12.

## Why
There should be a user friendly image on top of the buttons/grid so users can more easily identify where they clicked and specify this to the other party.

## How
You can either run the Cucumber tests or change the start activity set in `EntryActivity` which by default is set to `LoginActivity` and should be changed to `VisualVerificationActivity`.

## Alternative implementation
None considered

## Notes
n/a


app/src/main/res/layout/visual_verification.xml
```diff
@@ -0,0 +1,155 @@
...
+    </RelativeLayout>
+
+    <LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+        android:id="@+id/first_row"
+        android:orientation="horizontal"
+        android:layout_width="match_parent"
+        app:layout_heightPercent="25%">
+
+        <Button
+            android:id="@+id/button00"
```
jverbraeken: The names of all Android layout elements are global, so could you make the names of these buttons are little bit more specific like grid_verification_button00 or something like that?

clenrock:
I just pushed on your branch a small fix. There was an error where the contacts list was not set in the OnCreate method in the MainActivity, probably introduced while merging.

clenrock:
@ericcornelissen Is it possible to add grid lines to the Activity? (Probably just a PNG with the lines and the rest transparent)

app/src/androidTest/assets/features/visualVerification.feature
```diff
@@ -13,6 +13,6 @@ Feature: VisualVerification
     Examples:
       | first_button  | second_button | third_button | fourth_button | fifth_button |
       | 1             | 3             | 2            | 5             | 9            |
-      | 13            | 1             | 3            | 10            | 1            |
+      | 9             | 1             | 3            | 10            | 1            |
       | 2             | 4             | 2            | 4             | 2            |
-      | 2             | 15            | 7            | 14            | 8            |
```
TheBonheurs: why change these lines?

ericcornelissen: Because the size of the grid has been reduced 😄

PR #151, opened 2017-05-21
Test the service locator

jverbraeken:
- Relevant Issues: -
- Related Pull Requests: -

* * *

## What
This Pull Request adds to the repository tests for the service locator and excludes untestable modules like the activities

## Why
This Pull Request is needed because the service locator is the core of the app and should be tested throughly

## How
This feature can be viewed/tested within the project by running the tests

## Alternative implementation

## Notes

The test coverage is now 79%


app/src/test/java/com/nervousfish/nervousfish/service_locator/ServiceLocatorTest.java
```diff
@@ -0,0 +1,65 @@
...
+        serviceLocator = new ServiceLocator("foo");
+    }
+    @Test
+    public void testGetAndroidFilesDir() {
+        assertTrue("foo".equals(serviceLocator.getAndroidFilesDir()));
+    }
+
+    @Test
+    public void testGetDatabase() {
+        assertNotNull(serviceLocator.getDatabase());
```
TheBonheurs: Only assertNotNull tests? I think you can do better @jverbraeken

jverbraeken: Actually, I don't think I can. I don't care what the service locator chooses as implementation for the interface, I just want to be sure that there is an implementation instead of null. Of course I can test if a specific instance is returned, but that's not the point of tests at all! Every module, the service locator, should be a black box so the tests should only test the behaviour of this black box, not the very exact classes that are used in the implementation. Also, who watches the watchers?

app/fooaccountInformation.json
```diff
@@ -0,0 +1 @@
+[]
```
ericcornelissen: Can you remove this file 😄

app/foocontacts.json
```diff
@@ -0,0 +1 @@
+[]
```
ericcornelissen: Can you remove this file 😄

app/src/test/java/com/nervousfish/nervousfish/service_locator/ServiceLocatorTest.java
TheBonheurs: But where do you test if the implementation is right? Not in the module tests I think as there we look at the functionality not the return value of the service locator. Of course if the implementation works it kinda signifies that the service locator did a good job but you are never certain...

PR #138, opened 2017-05-18
Make a service of QRGenerator

jverbraeken:
- Relevant Issues: -
- Related Pull Requests: -

* * *

## What
This Pull Request adds to the repository the transfer of QRGenerator to the newly created util service

## Why
This Pull Request is needed because classes that offer a platform-dependent service should be encapsulated in a service to reduce the coupling between the application and the platform

## How
This feature can be viewed/tested within the project by checking if things are still working; running QRGeneratorTest

## Alternative implementation
-

PR #156, opened 2017-05-23
Refractor the qr generator to a service

jverbraeken:
- Relevant Issues: -
- Related Pull Requests: -

* * *

## What
This Pull Request adds to the repository the transfer of QRGenerator to the newly created util service

## Why
This Pull Request is needed because classes that offer a platform-dependent service should be encapsulated in a service to reduce the coupling between the application and the platform

## How
This feature can be viewed/tested within the project by checking if things are still working; running QRGeneratorTest

## Alternative implementation
-

## Notes

This is a copy of #138 , but this one doesn't include the "garbage" that the other PR had


app/src/main/java/com/nervousfish/nervousfish/modules/util/AUtils.java
```diff
@@ -0,0 +1,23 @@
+package com.nervousfish.nervousfish.modules.util;
+
+/**
+ * This class should be extended by all platform-specific {@link IUtils} implementations and contains
+ * all cross-platform functionality
+ */
+@SuppressWarnings("PMD.AbstractClassWithoutAnyMethod")
+abstract class AUtils implements IUtils {
```
ericcornelissen: Why does this Abstract Utils class only do something with QR, it isn't really a "utility" class, right?

app/src/main/java/com/nervousfish/nervousfish/modules/util/AndroidUtils.java
```diff
@@ -0,0 +1,55 @@
...
+ */
+public final class AndroidUtils extends AUtils {
+    private static final Logger LOGGER = LoggerFactory.getLogger("EncryptorAdapter");
+
+    /**
+     * Prevents construction from outside the class.
+     *
+     * @param serviceLocator Can be used to get access to other modules
+     */
+    @SuppressWarnings("PMD.UnusedFormalParameter")
```
ericcornelissen: The parameter is used, right?

app/src/main/java/com/nervousfish/nervousfish/modules/util/ZXingQRUtils.java
```diff
@@ -0,0 +1,106 @@
...
+import java.io.InvalidClassException;
+import java.util.EnumMap;
+
+/**
+ * A class providing all QR related functionality using Google's ZXing library
+ */
+final class ZXingQRUtils implements IQRUtils {
+    private static final Logger LOGGER = LoggerFactory.getLogger("ZXingQRUtils");
+    private final int qrcodeImageWidth;
+    private final int qrcodeImageHeight;
```
ericcornelissen:
```java
private final int qrCodeImageWidth;
private final int qrCodeImageHeight;
```

app/src/main/java/com/nervousfish/nervousfish/modules/util/AUtils.java
jverbraeken: Yes, but in the future we may add more utility functions to it. The alternative was to add a QRGenerator to the service locator, but when we get more utility classes the service locator will have a lot of methods. However, I just thought that some classes like the key generator are also what you could call "utility classes". Do you think that we should add a utility class as is done in this PR to the service locator and also refractor classes like th key generator to that class, or is it better to make every utility a separate entry in the service locator?

ericcornelissen: I would say every utility class should be added separately to the service locator 🙂

PR #158, opened 2017-05-23
Uml, closes #124 and closes #125 and closes #126

### LINKED ISSUE #124: EER diagram
- Parent issues: -
- Relevant issues: -
- Estimated effort: -

* * *

## User story
As a developer I want to have a high level overview of my database so that I can easily see what the current design of the database is.

## What
This issue demands to design an EER diagram

## Why
This issue is needed for the project because it's unclear for me as a developer how the database is designed.

## How
This feature can be added by drawing the EER diagram

## Definition of done

- An EER diagram representing the database is finished

### LINKED ISSUE #125: ER diagram
- Parent issues: -
- Relevant issues: -
- Estimated effort: -

* * *

## User story
As a developer I want to have a low level overview of my database so that I can easily see how the database is mapped onto the current DBMS.

## What
This issue demands to design an ER diagram

## Why
This issue is needed for the project because it's unclear for me as a developer how the database is mapped onto the DBMS.

## How
This feature can be added by drawing the ER diagram

## Definition of done

- An ER diagram representing the database is finished

### LINKED ISSUE #126: Use case diagram
- Parent issues: -
- Relevant issues: -
- Estimated effort: -

* * *

## User story
As a stakeholder I want to have an easy overview of what the developers think I want from the system so that I can be sure that the correct system is designed

## What
This issue demands a use case diagram

## Why
This issue is needed for the project because stackholders don't have a mimic of the real world and a view to easily understand how the system is designed yet.

## How
This feature can be added by drawing a use case diagram

## Definition of done

- The use case diagram is finished

jverbraeken:
- Relevant Issues: #124 #125 #126 
- Related Pull Requests: -

* * *

## What
This Pull Request adds to the repository some UML diagrams to make the implementation more clear to the stakeholders

## Why
This Pull Request is needed because it's very hard to understand the code just by looking at it

## How
This feature can be viewed/tested within the project by looking at the UML diagrams

## Alternative implementation

## Notes

This PR is mainly for reviewing. It's no problem if the diagrams are merged into develop, but the main purpose is to get feedback on these diagrams that will be included in the report ultimately.


TheBonheurs:
- I think while the first diagram may be obvious for us developers, it may be handy for our customer to see and have this document as a validation of us doing a good job.
- DB is very nice yes
- I thought @MrHug said that it would be handy to include de/encryption but I'm indeed not sure if should let it be used through the api like @clenrock said

jverbraeken:
@ericcornelissen 

> Got two initial comments:
> - Why is the first diagram useful? Isn't this (nearly) the same for all Android apps?
> - I do not fully agree with your last diagram. First of all it should be possible for third party apps to just get the public key of a contact so they can do the encryption themselves and I'm not even sure if we're supposed to do any encryption (although I do think that would be a nice addition).
> The same goes for decrypting the message, I think it should also be possible for third party apps to use the private key for the decryption themselves.

I thought the same as @TheBonheurs , that it's obvious for us developers, but that it may be handy for our customers to see how the deployment works. It's no complicated diagram indeed.
I don't agree with your last point, because your private key must be stored securely. Preferably even the user doesn't know it, because there is a lot of power in it. Why would a third-party application want to steal the very very secure private key of the user so that it can decrypt a message itself instead of just passing the encrypted message to our application and letting our application decrypt it?

jverbraeken:
@clenrock 

> The UseCaseDiagram is very clear for the first part. I think you can include the "Nice-to-haves" in the first part. And for the API, I think it should only have the ability to retrieve contacts or your own profile

Implemented!

jverbraeken:
@TheBonheurs We'll talk about it next meeting ;-)

PR #160, opened 2017-05-24
Let the backbutton exit the application

jverbraeken:
- Relevant Issues: -
- Related Pull Requests: -

* * *

## What
- exit after the back button is pressed in the loginactivity
- double tap the backbutton to exit in the mainactivity

## Why
This Pull Request is needed because you went back to the previous activity (possibly the entryactivity) in the old version.

## How
This feature can be viewed/tested within the project by using the back button in login / main activity

## Alternative implementation


ericcornelissen:
> I'm not sure if we actually want to use the Loginscreen as you say. Actually, it's not a loginscreen because you don't login but just fill in a password to verify your identity. When you've verified your identity, I don't think that anyone would ever want to go back to that "verify identity" screen to type again his password.

I don't really agree... You can go 2 ways with having to login to the app:
- Either you (nearly) always have to login when you open the app, think like a banking app or Paypal.
- Or you only have to login once, similar to most apps (e.g. Facebook, Twitter or Youtube).

However this second approach is not for us as it would kinda require a central server, otherwise the login  screen is completely useless. With your (i.e. the current) approach I would even say logging in is only a nuisance as it may be confusing to sometimes having to login and sometimes not having to.

I personally would go for the banking approach of logging in...

app/src/main/res/layout/activity_main.xml
```diff
@@ -41,6 +41,7 @@
         android:layout_height="wrap_content"
         android:layout_gravity="bottom|end"
         android:layout_margin="@dimen/fab_margin"
+        android:onClick="onClickFab"
```
TheBonheurs: FAB = Floating Action Button right? not sure if it's the most understandable name :p

ericcornelissen:
Thanks for listening to my suggestion ❤️ 

Unfortunately it doesn't seem to be working properly... See the `.gif` below:
backbutton

jverbraeken:
@ericcornelissen @TheBonheurs Implemented your suggestions!

app/src/main/java/com/nervousfish/nervousfish/activities/MainActivity.java
```diff
@@ -94,19 +93,61 @@ protected void onCreate(final Bundle savedInstanceState) {
...
+        startActivity(intent);
     }
 
     /**
      * Switches the sorting mode.
+     *
+     * @param view The sort floating action button that was clicked
      */
-    public void onSortButtonClick(final View view) {
+    public void onSortButtonClicked(final View view) {
```
ericcornelissen: Could you be consistent with naming these methods, sometimes you use `onClick` and sometimes `onClicked`...

I prefer `onClick` because it is active and shorter then `onClicked`. Also I think we've used `onClick` so far in other classes as well...

app/src/main/java/com/nervousfish/nervousfish/activities/MainActivity.java
```diff
@@ -94,19 +93,61 @@ protected void onCreate(final Bundle savedInstanceState) {
...
+    /**
+     * Gets triggered when the QR button is clicked.
+     *
+     * @param view - the ImageButton
+     */
+    public void onQRButtonClicked(final View view) {
+        LOGGER.info("QR button clicked");
+        final Intent intent = new Intent(this, QRActivity.class);
+        intent.putExtra(ConstantKeywords.SERVICE_LOCATOR, this.serviceLocator);
+        startActivity(intent);
```
ericcornelissen: Would you mind adding `this` before each `startAcitivty(intent)`? 😇

jverbraeken:
@ericcornelissen Fixed!

PR #167, opened 2017-05-26
Exclude untestable Bluetooth activities

jverbraeken:
- Relevant Issues: -
- Related Pull Requests: -

* * *

## What
This Pull Request adds to the repository jacoco exclude rules for the untestable Bluetooth activities

## Why
This Pull Request is needed because our test coverage doesn't give us an overview of what percentage of the classes is tested that can be tested otherwise

## How
This feature can be viewed/tested within the project by running jacoco

## Alternative implementation


config/coverage.gradle
```diff
@@ -5,16 +5,19 @@ task jacocoTestReport(type: JacocoReport, dependsOn: ['testDebugUnitTest', 'crea
         csv.enabled false
         html.enabled true
         html.destination "${buildDir}/reports/jacoco/test/html"
-        xml.enabled true
-        xml.destination "${buildDir}/reports/jacoco/test/jacocoTestReport.xml"
+        xml.enabled false
```
ericcornelissen: Could you re-enable the xml report (with the same destination), Coveralls depends on it 😇

PR #180, opened 2017-05-30
Diagram to show master slave interaction while pairing

clenrock:
- Relevant Issues: #104
- Related Pull Requests: -

* * *

## What
This Pull Request adds to the repository the documentation of the interaction between the master and slave while pairing with the extra identification method.

## Why
This Pull Request is needed because it gives more insight into the idea of pairing via Bluetooth with the extra identification method.

## Alternative implementation
We now use a method where we send the contact to both sides at the same time. This also doesn't need the sending of the tap data because the contact send is encrypted with this tapdata.


clenrock:
@ericcornelissen Made an update based on your suggestions

PR #181, opened 2017-05-30
Fix app not progressing after LoginActivity due to Bluetooth

ericcornelissen:
- Relevant Issues: -
- Related Pull Requests: -

* * *

## What
The app on `develop` currently doesn't work when the emulator/device has no Bluetooth or Bluetooth not enabled, this is fixed with this PR.

## Why
We want the app on develop to be working as good as it can, so it is kinda important that you can use the app further then just the `LoginActivity`.

## How
Open the app on an emulator, login and see for yourself that it works.

## Alternative implementation
None considered, a working app is more important for now 😅 

## Notes
n/a


app/src/main/java/com/nervousfish/nervousfish/activities/MainActivity.java
```diff
@@ -92,7 +93,16 @@ protected void onCreate(final Bundle savedInstanceState) {
 
         sortOnName();
 
-        this.serviceLocator.getBluetoothHandler().start();
+        try {
+            this.serviceLocator.getBluetoothHandler().start();
+        } catch (NoBluetoothException e) {
+            LOGGER.info("Bluetooth not available on device, hiding button");
```
jverbraeken: Can you show an information dialog to inform the user about this?

app/src/main/java/com/nervousfish/nervousfish/activities/MainActivity.java
```diff
@@ -92,7 +93,16 @@ protected void onCreate(final Bundle savedInstanceState) {
...
+            final View v = this.findViewById(R.id.pairing_menu_bluetooth);
+            v.setVisibility(View.GONE);
+        } catch (IOException e) {
+            LOGGER.info("Bluetooth handler not started, most likely Bluetooth is not enabled");
```
jverbraeken: Could you show a dialog to inform the user about this and ask him if Bluetooth should be turned on?
This can help you: https://stackoverflow.com/questions/3806536/how-to-enable-disable-bluetooth-programmatically-in-android

app/src/main/java/com/nervousfish/nervousfish/exceptions/NoBluetoothException.java
```diff
@@ -0,0 +1,24 @@
+package com.nervousfish.nervousfish.exceptions;
+
+/**
+ * Simple {@link RuntimeException} that is used to notify usage of Bluetooth is not avaiable.
+ */
+public class NoBluetoothException extends RuntimeException {
```
jverbraeken: Go to Settings -> Editor -> Inspections -> Java and enable everything from Class metrics, from Code style issues "Instance field access not qualified with this, Instance method not qualified with this", from Inheritance issues Missing Override annotation, everything from concurrency and for this line of code specifically: everything from serialization. When you hover over the classname our brilliant idea will hint you to generate a serialversionuid and nice readobject/writeobject methods.

app/src/main/java/com/nervousfish/nervousfish/activities/MainActivity.java
ericcornelissen: I don't think that is necessary, as they are probably aware they don't have Bluetooth, right?

app/src/main/java/com/nervousfish/nervousfish/exceptions/NoBluetoothException.java
ericcornelissen: And why do we want that? 😕

app/src/main/java/com/nervousfish/nervousfish/activities/MainActivity.java
ericcornelissen: Will do 👍

ericcornelissen: Done, see [link](https://github.com/ericcornelissen/NervousFish/pull/181/files#diff-149538cc1f1f5e4370b6ae93d6489330R108)

app/src/main/java/com/nervousfish/nervousfish/activities/MainActivity.java
jverbraeken: I don't think it's necessary as well, but if you put
`new SweetAlertDialog(this, SweetAlertDialog.WARNING_TYPE)
                    .setTitleText(getString(R.string.foo))
                    .setContentText(getString(R.string.bar))
                    .setConfirmText(getString(R.string.dialog_ok))
                    .setConfirmClickListener(null)
                    .show();`
in it, then the user doesn't even have to remember that he doesn't have Bluetooth

app/src/main/java/com/nervousfish/nervousfish/exceptions/NoBluetoothException.java
jverbraeken: class metrics, because you can't run SIG locally. instance field access / methods not qualified with this because you like to use this apparently. Everything from concurrency because it really are good practices and Checkstyle, PMD and Findbugs don't know anything about those. Everything from serialization, because serialversionuid: (copied from the documentation)

> The serialization runtime associates with each serializable class a version number, called a serialVersionUID, which is used during deserialization to verify that the sender and receiver of a serialized object have loaded classes for that object that are compatible with respect to serialization. If the receiver has loaded a class for the object that has a different serialVersionUID than that of the corresponding sender's class, then deserialization will result in an  InvalidClassException. If a serializable class does not explicitly declare a serialVersionUID, then the serialization runtime will calculate a default serialVersionUID value for that class based on various aspects of the class, as described in the Java(TM) Object Serialization Specification. **However, it is strongly recommended** that all serializable classes explicitly declare serialVersionUID values, since the default serialVersionUID computation is highly sensitive to class details that may vary depending on compiler implementations, and can thus result in unexpected InvalidClassExceptions during deserialization. Therefore, to guarantee a consistent serialVersionUID value across different java compiler implementations, a serializable class must declare an explicit serialVersionUID value. 

Also, http://index-of.es/JAVA/Effective%20Java.pdf , very bottom page 301. The readobject/writeobject are needed to that we can extend the exception with new fields in a later release. Great explanation at the beginning of page 299, http://index-of.es/JAVA/Effective%20Java.pdf

app/src/main/java/com/nervousfish/nervousfish/activities/MainActivity.java
ericcornelissen: That's why I'm hiding the Bluetooth button in the connection options

app/src/main/java/com/nervousfish/nervousfish/exceptions/NoBluetoothException.java
ericcornelissen: > instance field access / methods not qualified with this because you like to use this apparently.

Kind of a weird place to give this advice, but thanks anyway 😄 


> Everything from concurrency because it really are good practices...

Again, kinda of a weird place to mention this. Not getting any errors in this class from that...


> Everything from serialization, because serialversionuid...

I [added](https://github.com/ericcornelissen/NervousFish/pull/181/files#diff-a8926c7e0f9739e6fb5fed5d79dedefdR8) a `serialVersionUID` but am not convinced of adding a default `readObject` & `writeObject` method...

PR #170, opened 2017-05-27
Refactor Bluetooth

jverbraeken:
- Relevant Issues: -
- Related Pull Requests: -

* * *

## What
This Pull Request adds to the repository a full refractoring of the Bluetooth related classes.

- The EventBus is used extensively to communicate 
- A background service is introduced
- A global Application class is introduced to keep track of the background service
- Every data type can now be sent over Bluetooth (so not just contacts)
- The large Bluetooth classes is split in 5 new classes (and also for a part in APairingHandler)

## Why
This Pull Request is needed because the old Bluetooth class was difficult to understand, the Bluetooth screen had to be opened to receive contacts and only contacts could be sent and PMD didn't like the coupling / class size

## How
This feature can be viewed/tested within the project by sharing contacts over Bluetooth

## Alternative implementation
Other implementations that I've have considered are -

## Notes

The very first meeting I'll give you a walkthrough through this new code ;-)

PR #185, opened 2017-05-30
Add several tests for more code coverage

jverbraeken:
- Relevant Issues: #
- Related Pull Requests: #169 #171 #181 

* * *

## What
This Pull Request adds to the repository the tests that are needed to get maximum code coverage in combination with the tests in a another local branch that I'll push as soon as #181, #169 and #171 are merged

## Why
This Pull Request is needed because we want to know for sure that everything works correctly and tests are a nice tool to verify that

## How
This feature can be viewed/tested within the project by running the tests

## Alternative implementation
Other implementations that I've have considered are

PR #190, opened 2017-05-31
Added extra classes to include in the data wrapper#189

TheBonheurs:
- Relevant Issues: #170 
- Related Pull Requests: -

* * *

## What
Two extra classes that encapsylate collections of other classes.

## Why
This Pull Request is needed because 
We use a DataWrapper when we are sending data over bluetooth/nfc/qr and it encapsulates an object but won't work with a List. These new classes will save those lists within them.

## How
This feature can be viewed/tested within the project by 
Adding code to send multuple contacts or taps and seeing if they are parsed correctly.

## Alternative implementation
Using an arraylist (or any other list) as a param for DataWrapper but it proved to be cumbersome and produced nasty code...

## Notes

Not really


TheBonheurs:
Sure! I was looking at the Contact tests as reference and it was missing there though 😨

jverbraeken:
> Sure! I was looking at the Contact tests as reference and it was missing there though 😨 

Oops, fixed that in #185

app/src/main/java/com/nervousfish/nervousfish/data_objects/tap/ATapData.java
```diff
@@ -33,4 +33,29 @@ public Timestamp getTimestamp() {
...
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        
+        final ATapData ts = (ATapData) o;
+
+        return this.timestamp == null ? ts.timestamp == null : this.timestamp.equals(ts.timestamp);
```
ericcornelissen: Could you make this `equals` more like the one [in multicontacts](https://github.com/ericcornelissen/NervousFish/pull/190/files#diff-149fcabef3cefb35343ebbd60e7f63aeR38) 😇

TheBonheurs: Can you elaborate a bit more on this? What exactly should I change?

TheBonheurs: In what way do you want it to change? Which part exactly?

ericcornelissen: Something like:
```java
public boolean equals(final Object o) {
    if (o == null || this.getClass() != o.getClass()) {
        return false;
    }
        
    final ATapData that = (ATapData) o;
    return this.timestamp == null ? that.timestamp == null : this.timestamp.equals(that.timestamp);
}
```

PR #157, opened 2017-05-23
Add ActivateBluetoothActivity to activate bluetooth before users go to the Connection page

clenrock:
- Relevant Issues: #143 
- Related Pull Requests: #141 

* * *

## What
This Pull Request adds a new activity which will prompt the user to activate Bluetooth before they go to the actual connections page.

## Why
Before, if the user didn't have their Bluetooth on when they went to the connections page, it would try to load paired devices and discovered devices, but Bluetooth first have to be put on. But when it's on, you have to reload the page by going back and going to the page again. This makes it more user-friendly and also adds the message that we won't do anything with your location.

## How
This feature can be viewed/tested within the project by opening the app (on a real phone) and pressing the Bluetooth icon in the main activity.

## Alternative implementation
Other implementations that I've have considered are prompting the user to activate Bluetooth before going to the Connection scene. But this looked like the app was crashing.

## Notes
Note that I didn't create any tests yet, and that this doesn't close issue #143. The tests and the rest of the user-friendliness will be created when #141 is merged, that PR fixes some Bluetooth issues.

PR #194, opened 2017-06-01
Use popup instead of a new activity to activate bluetooth

clenrock:
- Relevant Issues: #157 
- Related Pull Requests: -

* * *

## What
This Pull Request deletes the ActivateBluetoothActivity and replaces it with a neat popup asking the user if he wants to enable Bluetooth.

## Why
This Pull Request is needed because this makes using Bluetooth much more user friendly and convenient. 

## How
This feature can be viewed/tested within the project by opening the app and pressing the Bluetooth button when Bluetooth is off.


app/src/main/java/com/nervousfish/nervousfish/activities/MainActivity.java
```diff
@@ -316,10 +303,40 @@ public void onClick(final SweetAlertDialog sDialog) {
...
+
+                        /**
+                         * {@inheritDoc}
+                         */
+                        @Override
+                        public void onClick(final SweetAlertDialog sweetAlertDialog) {
+                            sweetAlertDialog.dismiss();
+                            LOGGER.info("Requesting to enable Bluetooth");
+                            final Intent intent = new Intent(BluetoothAdapter.ACTION_REQUEST_ENABLE);
+                            startActivityForResult(intent, MainActivity.REQUEST_CODE_ENABLE_BLUETOOTH);
```
ericcornelissen: I would suggest to remove the `bluetoothButtonPressed` flag and replace it with additional data in this intent so that `onActivityResult` can detect it with the `data` parameter

config/quality/checkstyle/checkstyle.xml
```diff
@@ -50,7 +50,9 @@
         <module name="EmptyStatement" />
         <module name="EqualsAvoidNull" />
         <module name="EqualsHashCode" />
-        <module name="ExplicitInitialization" />
+        <module name="ExplicitInitialization">
```
TheBonheurs: Why actually?

config/quality/pmd/pmd-ruleset.xml
```diff
@@ -51,7 +51,9 @@
         <exclude name="ShortClassName" />
         <exclude name="VariableNamingConventions" />
     </rule>
-    <rule ref="rulesets/java/optimizations.xml" />
+    <rule ref="rulesets/java/optimizations.xml">
+        <exclude name="RedundantFieldInitializer" />
```
TheBonheurs: What does this option do?

app/src/main/java/com/nervousfish/nervousfish/activities/MainActivity.java
```diff
@@ -194,20 +178,23 @@ public void onPairingButtonClicked(final View view) {
         final Intent intent;
         switch (view.getId()) {
             case R.id.pairing_menu_bluetooth:
-                intent = new Intent(this, ActivateBluetoothActivity.class);
+                this.bluetoothButtonPressed = true;
+                this.enableBluetooth();
```
jverbraeken: Bluetooth should be enabled as soon as the user logs in, but we'll have to talk about how we want do do that (eg ask the user if he wants that). Made an issues for that: #198

config/quality/pmd/pmd-ruleset.xml
ericcornelissen: This rule complained about [this line](https://github.com/ericcornelissen/NervousFish/pull/194/files#diff-149538cc1f1f5e4370b6ae93d6489330R74). I don't think it is ever bad to explicitly initialize a variable...

config/quality/checkstyle/checkstyle.xml
jverbraeken: Because you can argue that `boolean foobar = false` or that `int baz = 0` is more clear and more explicit than just `boolean foobar` or `int baz` while it has no effect on the performance (the generated bytecode is exactly the same)

config/quality/pmd/pmd-ruleset.xml
jverbraeken: Exactly the same as Checkstyle's ExplicitInitialization, see my comment

app/src/main/java/com/nervousfish/nervousfish/activities/MainActivity.java
```diff
@@ -195,28 +154,30 @@ public void onPairingButtonClicked(final View view) {
         final Intent intent;
         switch (view.getId()) {
             case R.id.pairing_menu_bluetooth:
-                intent = new Intent(this, ActivateBluetoothActivity.class);
+                this.enableBluetooth(true);
```
ericcornelissen: I think it will be better to check and instantly go to the Bluetooth pairing screen within this method as opposed to the `enableBluetooth()` method

Issue #63, opened 2017-05-05
Import / export backup


jverbraeken:
- Parent issues: -
- Relevant issues: -
- Estimated effort: 8

* * *

## User story
As a user I want to be able to import / export my contact list with public keys so that I don't lose all my data when lose access to the app

## What
This issue demands the possibility to make a full backup of all data and to restore that backup

## Why
This issue is needed for the project because users switch from phones, phones crash, remove the app to free some space, etc. but they want to keep their data

## How
This feature can be added by adding buttons for importing / exporting a backup file and by attaching the correct functionality to those buttons.

## Definition of done
This task can be considered done when

- There is a button to import the backup
- There is a button to export the backup
- The import button lets the user choose where to import the backup from
- The export button lets the user choose where to export the backup to
- The data of the backup is reflected in the database after import
- The data of the database is contained in the backup after export

PR #199, opened 2017-06-02
Enable auto backup google#63

TheBonheurs:
- Relevant Issues: #63 
- Related Pull Requests: -

* * *

## What
Auto backup feature for apps provided by google

## Why
This Pull Request is needed because now we can assure the users their data will be backed up.

## How
This feature can be viewed/tested within the project by :

- The user has enabled backup on the device in Settings > Backup & Reset.
- At least 24 hours have elapsed since the last backup.
- The device is idle and charging.
- The device is connected to a Wi-Fi network. If the device is never connected to a wifi network, then Auto Backup never occurs.

## Alternative implementation
Other implementations that I've have considered are casting our contacts to google contacts and using that specific google contact backup feature but according to @jverbraeken it was too complicated.


ericcornelissen:
I have one important question about this feature, is it enabled by default?

I don't think our users, who are particularly nervous about their privacy, want to have this feature enabled by default... I would be okay with a checkbox in the first time use activity asking users if they want to use this feature.

app/src/main/java/com/nervousfish/nervousfish/modules/pairing/INfcHandler.java
```diff
@@ -5,5 +5,5 @@
 /**
  * Defines a module used for pairing using NFC that can be used by a service locator.
  */
-public interface INfcHandler extends IPairingHandler, IModule {
```
ericcornelissen: Why are you changing this in this PR?

app/src/main/java/com/nervousfish/nervousfish/modules/pairing/IQRHandler.java
```diff
@@ -6,5 +6,5 @@
  * Defines a module used for pairing using a QR code that can be used by a service locator.
  */
 
-public interface IQRHandler extends IPairingHandler, IModule {
```
ericcornelissen: Why are you changing this in this PR?

app/src/main/java/com/nervousfish/nervousfish/modules/pairing/INfcHandler.java
TheBonheurs: No idea, I've added them manually again...

app/src/main/java/com/nervousfish/nervousfish/modules/pairing/IQRHandler.java
TheBonheurs: Idem

TheBonheurs:
As to your question yes if you have Google backup enabled on your mobile device this will happen automatically. But I think if the users find this sketchy they will have that option disabled on their device. This will mean that their data won't be backed up though bu as I said in the alternatives I haven't found an elegant way to perform this backup manually.

ericcornelissen:
OK, just to be clear: If I don't have Google Backup enabled, then this feature won't do anything?

If that is the case I'll approve this PR 😄

TheBonheurs:
knipsel

ericcornelissen:
@TheBonheurs, why did you close this PR?

TheBonheurs:
No idea tbh @ericcornelissen , can't remember it either

TheBonheurs:
It could be that I moved this item on the waffle queue to done or something like that but it shouldn't immediately close this right?

ericcornelissen:
I think it might, it seems like a plausible explanation 😆

ericcornelissen:
Reopening this, so we can reconsider merging it 😄

PR #207, opened 2017-06-02
Skip the login page after "First use app" and "Create profile"

ericcornelissen:
- Relevant Issues: -
- Related Pull Requests: -

* * *

## What
This Pull Request updates the app flow so that the login activity is skipped the first time you use the app. 
- **Before**: When you first opened the app you had to specify your name/password and after that immediately login.
- **After**: When you first opened the app you had to specify your name/password and after that you go straight to the `MainActivity`.

## Why
It makes more sense to go directly to the `MainActivity`, you just specified your details anyways...

## How
Uninstall the app and install it again afterwards, you'll be prompted to create a profile and after that you should go straight to the `MainActivity`.

## Alternative implementation
None

## Notes
Note that using the back button in the `MainActivity` after you just created a profile still works, this is because I changed the `onBackButton()` code of the `MainActivity` to always return to the `LoginActivity`.


ericcornelissen:
> I find it better that you have to login after creating a profile

ericcornelissen:
> I find it better that you have to login after creating a profile

Why do you think that is better? If that is because it is that way in most (online) applications, the reason it is done there does not apply to our app. These applications require you to login after creating your account because you have to verify your email first, which is not the case for our application...

clenrock:
@ericcornelissen Fair point

app/src/main/java/com/nervousfish/nervousfish/activities/CreateProfileHelper.java
```diff
@@ -0,0 +1,129 @@
...
+     *
+     * @param keyType The type of key to generate.
+     * @return a {@link KeyPair} with the key type selected
+     */
+    KeyPair generateKeyPair(final IKey.Types keyType) {
+        switch (keyType) {
+            case RSA:
+                return this.keyGenerator.generateRSAKeyPair(CreateProfileHelper.DEFAULT_KEY_NAME);
+            case Simple:
+                final IKey publicKey = new SimpleKey("public", "foo");
```
jverbraeken: Wasn't simple key meant as a single string to both encrypt and decrypt a message instead of using a private and public key? Additionally, this code doesn't make sense, is hard-coded, I cannot imagine any reason why anybody would ever use it, so maybe it's better to replace simple key with DSA which would make a lot more sense imo.

ericcornelissen: > Wasn't simple key meant as a single string to both encrypt and decrypt a message instead of using a private and public key?

No it wasn't, it was meant as an example implementation of `IKey`


> I cannot imagine any reason why anybody would ever use it, so maybe it's better to replace simple key with DSA which would make a lot more sense imo.

True, but we don't have an implementation of DSA yet so I couldn't add a case for that.


I'll create a new issue for this so that we don't forget about it 😄

PR #195, opened 2017-06-01
Pairing via Bluetooth with the RhythmActivity

clenrock:
- Relevant Issues: #109 #104 #61 
- Related Pull Requests: -

* * *

## What
This Pull Request adds to the repository the integrated version of the RhythmActivity together with the whole flow of how the pairing will work.

## Why
This Pull Request is needed because first it were only separate parts, but now it's connected together.

## How
This feature can be viewed/tested within the project by pairing over Bluetooth.

## Alternative implementation
Other implementations that I've have considered are 

## Notes
BIG NOTE: this is not fully tested, the PR is opened for now to review the code, but currently we are testing still and trying to find errors which we can get out.

PR #208, opened 2017-06-02
Refactor / fix several bugs in the Bluetooth threads, closes #197

### LINKED ISSUE #197: Fix cancel methods in the Bluetooth threads
- Parent issues: -
- Relevant issues: -
- Estimated effort: -

* * *

## User story
As a developer I want that the cancel buttons in the Bluetooth threads cancel the thread so that threads can be cancelled to prevent unnecessary threads from existing

## What
This issue demands to let the cancel methods work but that they do only close the socket if they don't disconnect the user from his connection if that shouldn't be so

## Why
This issue is needed for the project because unnecessary threads have overhead and hard-to-solve consequences because they still do some work in the background

## How
This feature can be added by rewriting the cancel methods

## Definition of ready
This features is not ready until the following issues are finished:

- #195 

## Definition of done

- The cancel methods work but only close the socket if they don't disconnect the user from his connection if that shouldn't be so

jverbraeken:
- Relevant Issues: #197 
- Related Pull Requests: #195 

* * *

## What
- A boolean to the cancel method of AndroidBluetoothConnectThread that decides if the socket should be closed
- Makes EspressoTestMatchers in QRExchangeKeyActivitySteps static final, because findbugs complained
- A new exception, thrown when a socket could not be created
- Delegations to the Bluetooth threads so that the thread creator can exert better control over the thread's behavior and to better localize all concurrency related operations
- Thread synchronization on a lock instead of the object itself
- A better exception naming format by allowing exception names like e2

## Why
This Pull Request is needed because the socket of connectthread is never closed when the user is never connected, but only initiates the request.

## How
This feature can be viewed/tested within the project by verifying that Bluetooth still works.

## Alternative implementation

## Notes

~~#195 should be merged first~~


config/quality/checkstyle/checkstyle.xml
```diff
@@ -176,7 +176,9 @@
         <module name="AbstractClassName">
             <property name="format" value="^A[A-Z].+$" />
         </module>
-        <module name="CatchParameterName" />
+        <module name="CatchParameterName">
+            <property name="format" value="^[a-z][a-zA-Z0-9]*$" />
```
ericcornelissen: I agree with the change, but may I ask why you needed it?

app/src/main/java/com/nervousfish/nervousfish/modules/pairing/AndroidBluetoothService.java
```diff
@@ -42,20 +43,26 @@
...
     private static final Logger LOGGER = LoggerFactory.getLogger("AndroidBluetoothHandler");
     // Binder given to clients
-    private final IBinder mBinder = new LocalBinder();
-    private AndroidBluetoothConnectThread connectThread;
+    @SuppressWarnings("ThisEscapedInObjectConstruction")
+    private final IBinder mBinder = new AndroidBluetoothService.LocalBinder(this);
+    private final Object lock = new Object();
     private BluetoothState state = STATE_NONE;
+    @Nullable
+    private AndroidBluetoothConnectThread connectThread;
```
ericcornelissen: While you're working on this, may I ask why we need to have a separate variable for each Thread kind? Isn't it true that we can have only one of these threads running at any point in time?

app/src/main/java/com/nervousfish/nervousfish/modules/pairing/AndroidBluetoothConnectThread.java
```diff
@@ -44,54 +42,71 @@
...
-                socket.close();
+                this.socket.close();
                 LOGGER.warn("Socket closed!");
             } catch (final IOException e) {
                 LOGGER.error("Closing socket", e);
             }
         }
     }
+
+    private static final class ConnectThread implements Runnable {
```
ericcornelissen: Interesting change, is there any other reason for doing this besides hiding the abstraction?

config/quality/checkstyle/checkstyle.xml
jverbraeken: Because I like to give exceptions a naming like e2 and e3 if I don't care at all about which specific exception it is. The alternative was to use exception names like econtactExistedException which is too long (> 15 characters) which requires to change yet another checkstyle rule

ericcornelissen: 👍

app/src/main/java/com/nervousfish/nervousfish/modules/pairing/AndroidBluetoothConnectThread.java
jverbraeken: Yeah, because StackOverflow and Intellij suggest that it's better to implement Runnable rather than extend Thread, because you can only extend a single class but you can implement multiple other classes

clenrock:
Closed because it didn't work and we were unable to debug the cause

Issue #178, opened 2017-05-30
Settings screen to update own profile


clenrock:
- Parent issues: #27 
- Relevant issues: #162
- Estimated effort: 8 hours (Cornel estimation)

* * *

## User story
As a user I want to have a settings screen so that I can see what is in my profile and have the ability to change it.

## What
This issue demands a new activity where you can see and change your account information. Also, there should be a button in the MainActivity to go to this settings screen.

## Why
This issue is needed for the project because as we have a way to create a profile now, users probably want the ability to change it.

## Definition of ready
This features is not ready until the following issues are finished:

- #27 (Is already done as I'm writing this issue.)

## Definition of done
This task can be considered done when

- [x] the app has a seperate Activity for settings
- [x] there is the ability to change your name
- [x] there is the ability to change your password
- [x] there is the ability to add, change and delete public keys
- [x] you can go to the settings via a button in the MainActivity
- [ ] there are nice popups informing the user of invalid input or when the profile is saved successfully

PR #214, opened 2017-06-06
Settings screen with the ability to change your name

clenrock:
- Relevant Issues: #178 
- Related Pull Requests: -

* * *

## What
This Pull Request adds to the repository a settings screen with a lot of options. The ability to change your name is the only one that really works at the moment.

## Why
This Pull Request is needed because when you have a profile maybe in the future you would like to change it.

## How
This feature can be viewed/tested within the project by uninstalling the app, creating a new profile and then going to the settings screen via MainActivity.

## Alternative implementation
Now I've used the default Android implementation of the preferences, but I first tried to make a screen of my own, which became very ugly.

## Notes
Please give the look and the feel while going through the settings as much attention as the code.
This is an early PR, I will try to fix the errors still in the code today.


app/src/main/java/com/nervousfish/nervousfish/activities/AppCompatPreferenceActivity.java
```diff
@@ -0,0 +1,109 @@
...
+import android.support.v7.widget.Toolbar;
+import android.view.MenuInflater;
+import android.view.View;
+import android.view.ViewGroup;
+
+/**
+ * A {@link android.preference.PreferenceActivity} which implements and proxies the necessary calls
+ * to be used with AppCompat.
+ */
+public abstract class AppCompatPreferenceActivity extends PreferenceActivity {
```
jverbraeken: I hate the name of this activity, but unfortunately this is the way to go in Android

app/src/main/java/com/nervousfish/nervousfish/activities/SettingsActivity.java
```diff
@@ -0,0 +1,255 @@
...
+import java.util.List;
+
+/**
+ * A {@link PreferenceActivity} that presents a set of application settings. This
+ * is the place where your profile is changed.
+ */
+public class SettingsActivity extends AppCompatPreferenceActivity {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger("SettingsActivity");
+    private static String DISPLAY_NAME = "display_name";
```
jverbraeken: Either make this final or change it to camelCase

app/src/main/java/com/nervousfish/nervousfish/activities/SettingsActivity.java
```diff
@@ -0,0 +1,255 @@
...
+    private static IServiceLocator serviceLocator;
+    private static boolean FIRST_LOAD = true;
```
jverbraeken: same

app/src/main/java/com/nervousfish/nervousfish/activities/SettingsActivity.java
```diff
@@ -0,0 +1,255 @@
...
+
+    /**
+     * A preference value change listener that updates the preference's summary
+     * to reflect its new value.
+     */
+    private static Preference.OnPreferenceChangeListener sBindPreferenceSummaryToValueListener = new Preference.OnPreferenceChangeListener() {
+        @Override
+        public boolean onPreferenceChange(Preference preference, Object value) {
+            String stringValue = value.toString();
```
jverbraeken: log

app/src/main/java/com/nervousfish/nervousfish/activities/SettingsActivity.java
```diff
@@ -0,0 +1,255 @@
...
+
+            if (preference.getKey().equals(DISPLAY_NAME)) {
+                updateDisplayName(preference, stringValue);
```
jverbraeken: log

app/src/main/java/com/nervousfish/nervousfish/activities/SettingsActivity.java
```diff
@@ -0,0 +1,255 @@
...
+            } else if (preference instanceof ListPreference) {
+                // For list preferences, look up the correct display value in
```
jverbraeken: log

app/src/main/res/xml/pref_general.xml
```diff
@@ -0,0 +1,14 @@
+<PreferenceScreen xmlns:android="http://schemas.android.com/apk/res/android">
+    <CheckBoxPreference android:title="Remember Bluetooth verification method"
```
jverbraeken: enter

app/src/main/java/com/nervousfish/nervousfish/activities/SettingsActivity.java
```diff
@@ -0,0 +1,260 @@
...
+    private static volatile IServiceLocator serviceLocator;
+
+    /**
+     * A preference value change listener that updates the preference's summary
+     * to reflect its new value.
+     */
+    private static Preference.OnPreferenceChangeListener sBindPreferenceSummaryToValueListener = new Preference.OnPreferenceChangeListener() {
+        @Override
+        public boolean onPreferenceChange(final Preference preference, final Object value) {
+            final String stringValue = value.toString();
```
jverbraeken: log

app/src/main/java/com/nervousfish/nervousfish/activities/SettingsActivity.java
```diff
@@ -0,0 +1,260 @@
...
+
+            if (preference.getKey().equals(DISPLAY_NAME)) {
+                updateDisplayName(preference, stringValue);
```
jverbraeken: log

app/src/main/java/com/nervousfish/nervousfish/activities/SettingsActivity.java
jverbraeken: Either make this final or change it to camelCase

app/src/main/java/com/nervousfish/nervousfish/activities/SettingsActivity.java
jverbraeken: same

jverbraeken:
@clenrock I have no idea why my last two comments are outdated immediately and outdated immediately again after I changed them, but yeah, just that you know that those comments belong to my PR

clenrock:
@jverbraeken I pushed the same exact minute ;)

app/src/main/res/values/array.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <string-array name="verification_methods">
+        <item>Visual verification</item>
+        <item>Rhythm verification</item>
+        <item>Choose when pairing</item>
+    </string-array>
```
ericcornelissen: Why did you create a separate resource for this? Also (if you want to keep it a separate resource) give it a better name and do the same for the dutch translation :wink:

app/src/main/AndroidManifest.xml
```diff
@@ -49,22 +49,28 @@
...
         <activity
             android:name=".activities.QRExchangeKeyActivity"
             android:theme="@style/AppTheme.NoActionBar" />
-
-        <service android:name=".modules.pairing.AndroidBluetoothService" />
+        <activity
+            android:name=".activities.SettingsActivity"
+            android:label="@string/title_activity_settings" />
+        <activity
+            android:name=".activities.KeyManagementActivity" />
```
ericcornelissen: Can you keep it so that all the activities are listed before the services?

(Also, why did you change the enters in this file? In particular: L20, L25 and L75)

app/src/main/res/values/array.xml
clenrock: @ericcornelissen I think it's need to have this file if we would like to expand it with more static arrays. The name 'array.xml' is the default name for arrays, what would you give as a different name?

ericcornelissen: > I think it's needed to have this file if we would like to expand it with more static arrays.

In that case, I would suggest you do the same for [the dutch version of `strings.xml`](https://github.com/ericcornelissen/NervousFish/pull/214/files#diff-b0d05aebee027c15c77e36003f7bcadcR108)


> The name 'array.xml' is the default name for arrays, what would you give as a different name?

Seems a bit odd to me given .e.g string**s**.xml, but how does array**s**.xml sound?

app/src/main/AndroidManifest.xml
clenrock: I think that was done automatically because I use "Create Activity" when you press on a package, which automatically adds it to the manifest. I will change it

app/src/main/java/com/nervousfish/nervousfish/activities/AAppCompatPreferenceActivity.java
```diff
@@ -0,0 +1,161 @@
...
+
+    /**
+     * Returns the {@link AppCompatDelegate}.
+     * @return the {@link AppCompatDelegate}
+     */
+    private AppCompatDelegate getDelegate() {
+        if (mDelegate == null) {
+            mDelegate = AppCompatDelegate.create(this, null);
+        }
+        return mDelegate;
```
ericcornelissen: Why isn't this code in the constructor and don't you use `this.mDelegate` instead of `getDelegate()`?

app/src/main/java/com/nervousfish/nervousfish/activities/SettingsActivity.java
```diff
@@ -0,0 +1,264 @@
...
+                                : "");
+
+
+            } else {
+                LOGGER.info("Preference changed which is not a ListPreference, and not the display name");
+                // For all other preferences, set the summary to the value's
+                // simple string representation.
+                preference.setSummary(stringValue);
+            }
+            return true;
```
ericcornelissen: Are you sure you always want to return `true`?

app/src/main/java/com/nervousfish/nervousfish/activities/SettingsActivity.java
```diff
@@ -0,0 +1,264 @@
...
+        public boolean onPreferenceChange(final Preference preference, final Object value) {
+            LOGGER.info("Preference changed");
+            final String stringValue = value.toString();
+
+            if (preference.getKey().equals(DISPLAY_NAME)) {
+                LOGGER.info("Preference changed at the display name");
+                updateDisplayName(preference, stringValue);
+            } else if (preference instanceof ListPreference) {
+                LOGGER.info("Preference changed for a ListPreference");
+                // For list preferences, look up the correct display value in
```
ericcornelissen: Can you abstract this into its own implementation similar to [when the display name is changed](https://github.com/ericcornelissen/NervousFish/pull/214/files#diff-bf06f4c02cc485ea11575068ccada50eR53)

app/src/main/java/com/nervousfish/nervousfish/activities/SettingsActivity.java
```diff
@@ -0,0 +1,264 @@
...
+                        .getString(preference.getKey(), ""));
+    }
+
+    @Override
+    protected void onCreate(final Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        setupActionBar();
+
+        final Intent intent = getIntent();
+        if (serviceLocator == null) {
```
ericcornelissen: Why did you put in this if-statement?

app/src/main/java/com/nervousfish/nervousfish/activities/SettingsActivity.java
```diff
@@ -0,0 +1,264 @@
...
+
+    /**
+     * This method stops fragment injection in malicious applications.
+     * Make sure to deny any unknown fragments here.
+     *
+     * @param fragmentName The name of the fragment
+     * @return A boolean which is true when the fragment is valid
+     */
+    protected boolean isValidFragment(final String fragmentName) {
+        return PreferenceFragment.class.getName().equals(fragmentName)
```
ericcornelissen: Maybe put `PreferenceFragment.class.getName()` into a local variable?

app/src/main/java/com/nervousfish/nervousfish/activities/SettingsActivity.java
```diff
@@ -0,0 +1,264 @@
...
+        public void onCreate(final Bundle savedInstanceState) {
+            super.onCreate(savedInstanceState);
+            addPreferencesFromResource(R.xml.pref_general);
+            setHasOptionsMenu(true);
+
+            // Bind the summaries of EditText/List/Dialog/Ringtone preferences
+            // to their values. When their values change, their summaries are
+            // updated to reflect the new value, per the Android Design
+            // guidelines.
+            bindPreferenceSummaryToValue(findPreference("choose_verification_use_every_time"));
```
ericcornelissen: Change "choose_verification_use_every_time" into "choose_verification_method_every_time"

Also, maybe use a enum for this?

app/src/main/java/com/nervousfish/nervousfish/activities/SettingsActivity.java
```diff
@@ -0,0 +1,264 @@
...
+        }
+
+        @Override
+        public boolean onOptionsItemSelected(final MenuItem item) {
+            final int id = item.getItemId();
+            if (id == android.R.id.home) {
+                startActivity(new Intent(getActivity(), SettingsActivity.class));
```
ericcornelissen: Why don't you use `startActivity(new Intent(this, SettingsActivity.class));`?

app/src/main/res/values-nl/strings.xml
```diff
@@ -87,4 +87,21 @@
     <string name="show_qr_code">Toon QR code</string>
     <string name="contact_set_name">Geef contact een naam</string>
     <string name="contact_exists_message">Contact bestaat al, geef aub een unieke naam</string>
+
+    <!-- Strings related to Settings -->
+    <string name="title_activity_settings">Settings</string>
+    <string name="content_description_settings_button">Settings knop drie punten</string>
+
+    <!-- Example General settings -->
```
ericcornelissen: Why is this named with "Example"?

app/src/main/res/values-nl/strings.xml
```diff
@@ -87,4 +87,21 @@
...
+    <string name="pref_header_general">Algemeen</string>
+
+    <string name="pref_verification_method_to_use">Verificatie methode om te gebruiken</string>
+
+    <!-- Settings for Profile -->
+    <string name="pref_header_profile">Profiel informatie</string>
+
+    <string name="display_name">Naam</string>
```
ericcornelissen: What do you think about using "Publieke naam" here?

app/src/main/java/com/nervousfish/nervousfish/activities/SettingsActivity.java
clenrock: Because of extension I have to return a boolean, but I'm not sure if that should be true or false. In the example I used they also returned true after setting the summaries of the elements.

app/src/main/java/com/nervousfish/nervousfish/activities/SettingsActivity.java
clenrock: Because this method is also run when opening a fragment, and it should only be set the first time the SettingsActivity is created.

app/src/main/java/com/nervousfish/nervousfish/activities/SettingsActivity.java
clenrock: I'm not sure because it is only used once.

app/src/main/java/com/nervousfish/nervousfish/activities/SettingsActivity.java
ericcornelissen: Wierd, but OK 😕

app/src/main/java/com/nervousfish/nervousfish/activities/SettingsActivity.java
ericcornelissen: From the documentation ([link](https://developer.android.com/reference/android/preference/Preference.OnPreferenceChangeListener.html)):
> True to update the state of the Preference with the new value.


Would you be OK with putting a separate `return true;` in each branch to make sure you explicitly have to return if you want to set the setting? (although I'm not sure if our static analysis tools are OK with that 😓)

app/src/main/java/com/nervousfish/nervousfish/activities/SettingsActivity.java
ericcornelissen: Are you sure it is only used once?
```java
protected boolean isValidFragment(final String fragmentName) {
    return PreferenceFragment.class.getName().equals(fragmentName)
        || GeneralPreferenceFragment.class.getName().equals(fragmentName)
        || ProfilePreferenceFragment.class.getName().equals(fragmentName);
}
```

app/src/main/java/com/nervousfish/nervousfish/activities/SettingsActivity.java
clenrock: This can't be used because it's in a fragment, not an activity itself.

app/src/main/java/com/nervousfish/nervousfish/activities/SettingsActivity.java
clenrock: Yes

app/src/main/java/com/nervousfish/nervousfish/activities/SettingsActivity.java
clenrock: But what to return in the end? False or throw an error? I'm not sure I want to change this

app/src/main/java/com/nervousfish/nervousfish/activities/SettingsActivity.java
ericcornelissen: Seems a bit odd te me, but sure 😕 Would you mind splitting the code and adding comments? e.g.
```java
final Activity activity = this.getActivity(); // I'm guessing this returns an Activity
final Intent intent = new Intent(activity, SettingsActivity.class); // Needed bacause we're in a Fragment
this.startActivity(intent);
```

app/src/main/java/com/nervousfish/nervousfish/activities/SettingsActivity.java
ericcornelissen: I guess throw an error, if we test properly we should always catch that before release... I get your concern but I like being explicit in situations like these...

app/src/main/java/com/nervousfish/nervousfish/activities/SettingsActivity.java
```diff
@@ -219,7 +227,7 @@ public void onCreate(final Bundle savedInstanceState) {
             // to their values. When their values change, their summaries are
             // updated to reflect the new value, per the Android Design
             // guidelines.
-            bindPreferenceSummaryToValue(findPreference("choose_verification_use_every_time"));
+            bindPreferenceSummaryToValue(findPreference("choose_verification_method_every_time"));
```
ericcornelissen: Could you use either a `ConstantKeywords` or a `this.getString(R.strings.id)` here?

app/src/main/res/values-nl/strings.xml
```diff
@@ -97,20 +97,25 @@
...
-    <!-- Settings for Profile -->
+    <!-- Profile Settings -->
     <string name="pref_header_profile">Profiel informatie</string>
 
-    <string name="display_name">Naam</string>
+    <string name="display_name">Publieke naam</string>
 
     <!-- KeyManagementActivity -->
     <string name="pref_key_management_header">Key management</string>
+=======
```
ericcornelissen: Interesting addition 🤔

clenrock: Mmmm...

app/src/main/java/com/nervousfish/nervousfish/ConstantKeywords.java
```diff
@@ -15,6 +15,8 @@
     public static final String WAIT_MESSAGE = "Wait message";
     public static final String DATA_RECEIVED = "Data received";
     public static final String TAP_DATA = "Tap data";
+    public static final String CHOOSE_VERIFICATION_PREFERENCE = "choose_verification_method_every_time";
+    public static final String DISPLAY_NAME = "display_name";
```
ericcornelissen: Why are you not using the value in `strings.xml` for this?

clenrock: Because it's a "key" not a string

app/src/main/java/com/nervousfish/nervousfish/activities/SettingsActivity.java
clenrock: Sure, I'll change it

app/src/main/java/com/nervousfish/nervousfish/ConstantKeywords.java
ericcornelissen: What do you mean?

jverbraeken: Because the user will never see that string, it's only used internally because some Android libraries don't support enums for some reason and require you to use Strings to identify things.

app/src/main/java/com/nervousfish/nervousfish/activities/SettingsActivity.java
```diff
@@ -18,17 +18,23 @@
...
 
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import edu.umd.cs.findbugs.annotations.SuppressWarnings;
+
 /**
  * A {@link PreferenceActivity} that presents a set of application settings. This
  * is the place where your profile is changed.
  */
+@SuppressFBWarnings(value = "ST_WRITE_TO_STATIC_FROM_INSTANCE_METHOD")
+//Because of the structure with the static methods, a static attribute has to be set.
```
ericcornelissen: Can you elaborate on why we need the structure with static methods?

clenrock: This is the default usage of the settings screen. The only place where the static method is important is at the on preference change listener. This is one method which is triggered for all changes in preferences from the settings screen. Because this method uses the servicelocator, we have to set that in the onCreate.

I'm actually thinking that maybe we can statically get it from the NervousFish class?

ericcornelissen: > I'm actually thinking that maybe we can statically get it from the NervousFish class?

That is for certain true 😄 Also (if it still applies) can you add - a shorter version of - the description you gave here in the code 😇

PR #217, opened 2017-06-06
Add hardware acceleration

jverbraeken:
- Relevant Issues: -
- Related Pull Requests: -

* * *

## What
This Pull Request adds to the repository hardware acceleration for the application

## Why
This Pull Request is needed because it's good practice to use hardware acceleration when it's available because it's a lot faster

## How
This feature can be viewed/tested within the project by running the app and verifying that everything still works

## Alternative implementation

## Notes

I couldn't test if the app still works myself, so it's important that you check if before you approve this PR.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -9,6 +9,7 @@
 
     <application
         android:allowBackup="true"
+        android:hardwareAccelerated="true"
         android:fullBackupContent="true"
         android:icon="@mipmap/app_logo_round"
         android:label="@string/app_name"


TheBonheurs:
> This Pull Request is needed because it's **good practice** to use hardware acceleration **when it's available** because it's a lot faster

Do you have a source on this?

ericcornelissen:
From stackoverflow ([link](https://stackoverflow.com/questions/16432565/android-hardware-acceleration-to-use-or-not-to-use)):
> It is advised to use hardware acceleration only if you have complex custom computations for scaling, rotating and translating of images, but do not use it for drawing lines or curves (and other trivial operations) ([source](https://developer.android.com/training/custom-views/optimizing-view.html#accelerate))

jverbraeken:
@ericcornelissen I agree with you

Issue #216, opened 2017-06-06
Click on the text besides the fab button in the MainActivity


clenrock:
- Parent issues: - 
- Relevant issues: - #52 
- Estimated effort: - 2 hours (estimated by Cornel)

* * *

## What
When you click the text, for example "Bluetooth" or "NFC", which is besides the fab buttons on the MainActivity the app crashes.

## How
The error is pretty descriptive and says something like "IllegalArgumentException: Only existing buttons can be clicked". So this can be fixed pretty easily

## Definition of done
This task can be considered done when

- the app doesn't crash anymore when the text besides the button is pressed

PR #219, opened 2017-06-06
Make it possible to click on the label, fixes the bug #216

clenrock:
- Relevant Issues: #216 
- Related Pull Requests: -

* * *

## What
This Pull Request adds to the repository a fix for the bug described at #216.

## How
This feature can be viewed/tested within the project by opening the MainActivity and clicking on the fab button's label.

## Notes
Unfortunately I had to make the switch-case an if-else statement


ericcornelissen:
Have you considered looking at the `labelFor` property of the `Label` in order to get the ID of the actual button?

clenrock:
@ericcornelissen No, do you think that makes a nicer solution?

ericcornelissen:
If it works it would because you could keep the `switch`-statement

clenrock:
@ericcornelissen I tried it, but the ID that is returned is not the one of the button. So I'll keep it like this

ericcornelissen:
Too bad 😢

Issue #196, opened 2017-06-01
Bluetooth communication without hardcoded strings


jverbraeken:
- Parent issues: -
- Relevant issues: -
- Estimated effort: 2

* * *

## User story
As a developer I want to send data objects instead of raw strings over the network so that my types are type-safe and extensible

## What
This issue demands to replace the strings that are send over the network by appropriate data classes

## Why
This issue is needed for the project because the current approach is type-unsafe and inextensible

## How
This feature can be added by replacing the strings that are send over the network by appropriate data classes

## Definition of ready
This features is not ready until the following issues are finished:

- #195 

## Definition of done
This task can be considered done when

-  The strings that are sent over the network are replaced by appropriate data classes

PR #221, opened 2017-06-07
No hardcoded strings for sending which verification method to use

clenrock:
- Relevant Issues: #196 
- Related Pull Requests: -

* * *

## What
This Pull Request adds to the repository an extra class and an enumerator which doesn't use the hardcoded strings anymore.

## Why
This Pull Request is needed because this is neater and gives more extensible. 

## How
This feature can be viewed/tested within the project by looking at the code, when running the app it should work the same as normally.


ericcornelissen:
🎉

Issue #182, opened 2017-05-30
No servicelocator when restarting after crash


clenrock:
- Parent issues: -
- Relevant issues: -

* * *

## Description of the bug
When the app crashes and you press restart the app, the app will restart on the activity where it crashed. The problem here is that when it restarts at another place than the EntryActivity, the servicelocator is not created. So the app crashes again because the servicelocator is null.

## Possible solution
Maybe check if a message or event is shown that there was a restart, because this should go to the EntryActivity.

clenrock:
Maybe a fix would be to have a check in every onStart method of an Activity which checks if ServiceLocator is null. If that is true, then go to the EntryActivity.
(Could be that there is a nicer fix possible)

ericcornelissen:
> Maybe a fix would be to have a check in every onStart method of an Activity which checks if ServiceLocator is null. If that is true, then go to the EntryActivity.

That would be a nice hotfix imo, we can look at other solutions later 😄

PR #223, opened 2017-06-07
Make the ServiceLocator globally accessible via NervousFish

ericcornelissen:
- Relevant Issues: #182
- Related Pull Requests: -

* * *

## What
This Pull Request updates the code such that the `ServiceLocator` can be obtained via a call to NervousFish. First and foremost this solves issue #182, but it also makes using activities a little less of a pain since you're no longer required to put the `ServiceLocator` in an `Intent`.

## Why
This Pull Request is needed because we want to fix the issue in #182 (crash due to a crash), but also because the update will make using the `ServiceLocator` a bit easier.

## How
This feature can be viewed/tested within the project by just opening the app and trying stuff out 😄 

## Alternative implementation
Other implementations that I've have considered all failed one way or another. Android does not seem to be consistent with restarting and restarting after crashes which causes the ServiceLocator or services (e.g. Bluetooth) to not always be available. Now that all these things are in the `NervousFish` class (which does seem to always start) everything should be available...

## Notes
Make sure you really run the app and do something, this way we can catch potential issues before the code is in the `develop` branch.


app/src/main/java/com/nervousfish/nervousfish/service_locator/NervousFish.java
```diff
@@ -10,86 +10,84 @@
...
     @Override
     public void onCreate() {
         super.onCreate();
 
         final Intent serviceIntent = new Intent(this, AndroidBluetoothService.class);
-        bindService(serviceIntent, connection, Context.BIND_AUTO_CREATE);
+        this.bindService(serviceIntent, this.connection, Context.BIND_AUTO_CREATE);
 
+        final String androidFileDir = this.getFilesDir().getPath();
+        NervousFish.serviceLocator = new ServiceLocator(androidFileDir);
```
jverbraeken: Is there a way to use another service locator in the tests  if the need arises somewhere in the future? If not, then we must check explicitly if the device is capable of using Bluetooth, sounds, etc. However, just thinking, maybe it is very good practice to do that checking explicitly...

ericcornelissen: > Is there a way to use another service locator in the tests if the need arises somewhere in the future?

Not that I know/can think off..


> However, just thinking, maybe it is very good practice to do that checking explicitly

Agreed 😄 Also, we have some very basic checks for Bluetooth but I'm not sure how good that works...

ericcornelissen:
> What do you think of the idea to make an AModule that contains the servicelocator?

Not sure what exactly you mean by an `AModule`, but I don't think we need that kind of abstraction here... Since we use a ServiceLocator, which basically provides other classes with `IModules`, I cannot think of another class that would ever become an `AModule`.


> Now you have to get the service locator in every module from NervousFish. Apart from increasing coupling...

Good point, we could perhaps make the ServiceLocator itself `static` so that all the modules can be obtained everywhere directly, also removing the need for having an `IServiceLocator` attribute in each class?

app/src/main/java/com/nervousfish/nervousfish/activities/FirstUseActivity.java
```diff
@@ -19,22 +17,14 @@
...
     @Override
     protected void onCreate(final Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
         this.setContentView(R.layout.activity_first_use);
-
-        final Intent intent = this.getIntent();
-        this.serviceLocator = (IServiceLocator) intent.getSerializableExtra(ConstantKeywords.SERVICE_LOCATOR);
-
-        LOGGER.info("FirstUseActivity created");
+        LOGGER.info("activity created");
```
clenrock: Did you remove the name of the activity because the logger already shows that?

app/src/main/java/com/nervousfish/nervousfish/activities/MainActivity.java
```diff
@@ -86,9 +86,7 @@ protected void onCreate(final Bundle savedInstanceState) {
             LOGGER.error("Failed to retrieve contacts from database", e);
         }
 
-        sorter = new MainActivitySorter(this);
-        sorter.sortOnName();
```
clenrock: Don't we want to sort on name when this activity is created?

app/src/main/java/com/nervousfish/nervousfish/activities/WaitActivity.java
```diff
@@ -55,24 +52,33 @@ protected void onCreate(final Bundle savedInstanceState) {
...
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    protected void onStart() {
+        super.onStart();
+        this.serviceLocator.registerToEventBus(this);
+
+        if (this.dataReceived != null && this.tapCombination != null) {
+            evaluateData();
```
clenrock: Add this. ? (My fault I know)

app/src/main/java/com/nervousfish/nervousfish/activities/FirstUseActivity.java
clenrock: (Maybe capitalize)

app/src/main/java/com/nervousfish/nervousfish/activities/MainActivity.java
ericcornelissen: We do, but we also want to sort on name when the activity is started or resumed, that is why you can find `sorter.sortOnName()` in the [`onResume()` method](https://github.com/ericcornelissen/NervousFish/pull/223/files#diff-149538cc1f1f5e4370b6ae93d6489330R133)

You can see that this should work based on the [Android Activity Lifecycle diagram](https://developer.android.com/guide/components/images/activity_lifecycle.png) from the [Android documentation](https://developer.android.com/guide/components/activities/activity-lifecycle.html)

app/src/main/java/com/nervousfish/nervousfish/activities/FirstUseActivity.java
ericcornelissen: Indeed I did 🙂 This avoid incorrect logs when people copy the `onCreate` method to other activities

app/src/main/java/com/nervousfish/nervousfish/activities/WaitActivity.java
ericcornelissen: I might add it if anyone requests any changes 😉 Otherwise I'm not going to create a commit just for that

jverbraeken:
```
> Now you have to get the service locator in every module from NervousFish. Apart from increasing coupling...

Good point, we could perhaps make the ServiceLocator itself `static` so that all the modules can be obtained everywhere directly, also removing the need for having an `IServiceLocator` attribute in each class?
```
I don't really like to make things static. It's anti-OOP and it increases coupling a lot. If we ever need to use another service locator because we port everything to Windows or so, than we've a problem. What we can do is adding a public static ServiceLocator field in NervousFish and, because NervousFish.ServiceLocator is too much work to type, do
- rename NervousFish to App and/or rename the public static (final) (transient) field to SL
- static import the field. This is very common practice for for example the JUnit library so we can also use it for our project

ericcornelissen:
Making the `ServiceLocator` in `NervousFish` is not possible because it must be created in the `onCreate` method... 🙁

PR #231, opened 2017-06-08
Make it possible to exchange contacts via nfc, not yet fully manually tested #8

TheBonheurs:
- Relevant Issues: #8
- Related Pull Requests: -

* * *

## What
This pr adds a new activity which should be able to exchange contacts over nfc (Android Beam)

## Why
This Pull Request is needed because NFC is MITM-proof and requires less hassle to setup.

## How
This feature can be viewed/tested within the project by clicking the pairing menu in the main activity and selecting NFC. This should lead to the NFCActivity which should open up the activity with an instruction gif of what to do.

## Alternative implementation
Other implementations that I've have considered are making it so that our nfc handler fully replaces the android nfc adapter but that proved difficult.

## Notes

Yes, as it's stated in the title this functionality needs to be tested with actual phones, the activity (design wise) can still be improved and there are even some activity transitions missing but it will be add to this pr before merging.


app/build.gradle
```diff
@@ -58,6 +58,8 @@ dependencies {
     compile 'com.wang.avi:library:2.1.3'
     compile 'com.android.support:support-v4:25.3.1'
     compile 'com.android.support:support-vector-drawable:25.3.1'
+    compile 'com.github.bumptech.glide:glide:4.0.0-RC0'
+    compile 'com.android.support:support-v4:25.3.1'
```
jverbraeken: This is duplicate

app/src/main/java/com/nervousfish/nervousfish/activities/NFCActivity.java
```diff
@@ -1,22 +1,147 @@
...
 
 /**
- * An {@link Activity} that used for pairing over NFC
+ * An {@link Activity} that beams NDEF Messages to Other Devices.
  */
-public class NFCActivity extends Activity {
+public class NFCActivity extends Activity implements CreateNdefMessageCallback {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger("NFCActivity");
+    private TextView textView;
```
jverbraeken: Can you make this name a bit more specific?

app/src/main/java/com/nervousfish/nervousfish/activities/NFCActivity.java
```diff
@@ -1,22 +1,147 @@
...
+import java.util.Arrays;
+
+import static android.nfc.NdefRecord.createExternal;
 
 /**
- * An {@link Activity} that used for pairing over NFC
+ * An {@link Activity} that beams NDEF Messages to Other Devices.
  */
-public class NFCActivity extends Activity {
+public class NFCActivity extends Activity implements CreateNdefMessageCallback {
```
jverbraeken: final

app/src/main/java/com/nervousfish/nervousfish/activities/NFCActivity.java
```diff
@@ -1,22 +1,147 @@
...
-     * @param savedInstanceState The saved state of the instance.
+     * {@inheritDoc}
      */
     @Override
-    protected void onCreate(final Bundle savedInstanceState) {
+    public void onCreate(final Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
         setContentView(R.layout.activity_nfc);
+
+        final Intent intent = getIntent();
```
jverbraeken: Please download https://www.dropbox.com/s/wryiggq2x9yr0tz/Project_Default.xml?dl=0
Press Ctrl-alt-s -> Editor -> Inspections -> Manage -> Import
Hover over getIntent and press Cleanup code

app/src/main/java/com/nervousfish/nervousfish/activities/NFCActivity.java
```diff
@@ -1,22 +1,147 @@
...
+        final NfcAdapter nfcAdapter = NfcAdapter.getDefaultAdapter(this);
+        if (nfcAdapter == null) {
+            Toast.makeText(this, "NFC is not available", Toast.LENGTH_LONG).show();
+            finish();
+            return;
+        }
+        LOGGER.info("Start creating an NDEF message to beam");
+        Glide.with(this).load(R.drawable.s_contact_animado).into((ImageView) findViewById(R.id.nfc_gif));
+        try {
+            final Profile myProfile = this.serviceLocator.getDatabase().getProfiles().get(0);
```
jverbraeken: law of demeter

app/src/main/java/com/nervousfish/nervousfish/activities/NFCActivity.java
```diff
@@ -1,22 +1,147 @@
...
+            LOGGER.info("Sending my profile with name: " + myProfile.getName() + ", public key: "
```
jverbraeken: String concatenation is very, very slow. Use "asdfasdfasdf {} asdf {} asdf", myProfile.getName(), my.... instead

app/src/main/java/com/nervousfish/nervousfish/activities/NFCActivity.java
```diff
@@ -1,22 +1,147 @@
...
+                    + myProfile.getPublicKey().toString());
+            final Contact contact = new Contact(myProfile.getName(), new SimpleKey("simplekey", "73890ien"));
+            bytes = this.serviceLocator.getNFCHandler().objectToBytes(contact);
```
jverbraeken: law of demeter

app/src/main/java/com/nervousfish/nervousfish/activities/NFCActivity.java
```diff
@@ -1,22 +1,147 @@
...
+        } catch (IOException e) {
+            LOGGER.error("Could not serialize my contact to other device " + e.getMessage());
```
jverbraeken: Now you just throw the full stacktrace of e away, you can better use
`LOGGER.error("Could not serialize my contact to other device", e);`

app/src/main/java/com/nervousfish/nervousfish/activities/NFCActivity.java
```diff
@@ -1,22 +1,147 @@
...
         setContentView(R.layout.activity_nfc);
+
+        final Intent intent = getIntent();
+        this.serviceLocator = (IServiceLocator) intent.getSerializableExtra(ConstantKeywords.SERVICE_LOCATOR);
+
+        textView = (TextView) findViewById(R.id.nfc_instruction);
+        // Check for available NFC Adapter
+        final NfcAdapter nfcAdapter = NfcAdapter.getDefaultAdapter(this);
+        if (nfcAdapter == null) {
+            Toast.makeText(this, "NFC is not available", Toast.LENGTH_LONG).show();
```
jverbraeken: Can you change this to a nice SweetAlert? Press ctrl-alt-shift-f for a lot of examples

app/src/main/java/com/nervousfish/nervousfish/activities/NFCActivity.java
```diff
@@ -1,22 +1,147 @@
...
+            finish();
+            return;
+        }
+        LOGGER.info("Start creating an NDEF message to beam");
+        Glide.with(this).load(R.drawable.s_contact_animado).into((ImageView) findViewById(R.id.nfc_gif));
+        try {
+            final Profile myProfile = this.serviceLocator.getDatabase().getProfiles().get(0);
+            LOGGER.info("Sending my profile with name: " + myProfile.getName() + ", public key: "
+                    + myProfile.getPublicKey().toString());
+            final Contact contact = new Contact(myProfile.getName(), new SimpleKey("simplekey", "73890ien"));
```
jverbraeken: You can get the user's key from the database

app/src/main/java/com/nervousfish/nervousfish/modules/pairing/NFCHandler.java
```diff
@@ -0,0 +1,70 @@
...
+    @Override
+    public void send(final byte[] buffer) {
+        // The NFC Handler handles the exchange of bytes in the activity
+    }
+
+
+    @Override
+    public void dataReceived(final byte[] bytes) {
+        getDataReceiver().get().dataReceived(bytes);
+    }
```
jverbraeken: Please add this code:
`/**
     * Deserialize the instance using readObject to ensure invariants and security.
     *
     * @param stream The serialized object to be deserialized
     */
    private void readObject(final ObjectInputStream stream) throws IOException, ClassNotFoundException {
        stream.defaultReadObject();
        ensureClassInvariant();
    }

    /**
     * Used to improve performance / efficiency
     *
     * @param stream The stream to which this object should be serialized to
     */
    private void writeObject(final ObjectOutputStream stream) throws IOException {
        stream.defaultWriteObject();
    }

    /**
     * Ensure that the instance meets its class invariant
     *
     * @throws InvalidObjectException Thrown when the state of the class is unstbale
     */
    private void ensureClassInvariant() throws InvalidObjectException {
        // No checks to perform
    }`

app/src/main/java/com/nervousfish/nervousfish/modules/pairing/NFCHandler.java
```diff
@@ -0,0 +1,70 @@
...
+            oos.writeObject(new DataWrapper(object));
+            oos.flush();
+            bytes = bos.toByteArray();
+        }
+        return bytes;
+    }
+
+    @Override
+    public void send(final byte[] buffer) {
+        // The NFC Handler handles the exchange of bytes in the activity
```
jverbraeken: Can you throw an error and log it if it happens anyway?

app/src/main/java/com/nervousfish/nervousfish/modules/pairing/NFCHandler.java
```diff
@@ -0,0 +1,70 @@
...
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.ObjectOutputStream;
+import java.io.Serializable;
+
+/**
+ * An handler for NFC communication without implementation, needed because NFC is unavailable on the emulator
+ */
+@SuppressWarnings("PMD.SingularField") // it's the serviceLocator, which isn't being used yet
```
jverbraeken: serviceLocator is no field

app/src/main/res/layout/activity_nfc.xml
```diff
@@ -1,5 +1,26 @@
...
+
+    <ImageView
+        tools:ignore = "UseCompoundDrawables"
+        android:id="@+id/nfc_gif"
+        android:layout_width="match_parent"
+        android:layout_height="442dp"
+        android:contentDescription="@string/descr_gif"
+        android:src="@android:drawable/sym_def_app_icon" />
+
+    <TextView
```
jverbraeken: Can you center the text?

app/src/main/res/layout/activity_nfc.xml
```diff
@@ -1,5 +1,26 @@
...
-    android:layout_height="match_parent">
+    android:layout_height="match_parent"
+    android:weightSum="1"
+    tools:ignore = "UseCompoundDrawables">
+
+    <ImageView
+        tools:ignore = "UseCompoundDrawables"
+        android:id="@+id/nfc_gif"
+        android:layout_width="match_parent"
+        android:layout_height="442dp"
```
jverbraeken: This should be in dimens.xml

app/src/main/res/values-nl/strings.xml
```diff
@@ -97,6 +97,10 @@
     <string name="contact_set_name">Geef contact een naam</string>
     <string name="contact_exists_message">Contact bestaat al, geef aub een unieke naam</string>
 
+    <!-- NFC -->
+    <string name="descr_gif">Dit zou een gif moeten zijn die nfc connection uitbeeldt.</string>
+    <string name="nfc_explanation">Breng uw toestel zo dicht mogelijk bij het gewenste andere toestel om contacten uit te wisselen .</string>
```
jverbraeken: spatie voor de .

app/src/main/res/values/strings.xml
```diff
@@ -99,4 +99,9 @@
     <!-- WaitActivity -->
     <string name="wait_message_slave_verification_method">Waiting for the other device to choose a verification method…</string>
     <string name="wait_message_partner_rhythm_tapping">Waiting for your partner to finish tapping…</string>
+
+    <!-- NFC -->
+    <string name="descr_gif">This should be a gif representing nfc usage.</string>
```
jverbraeken: I think that "representing the use of NFC" is better. Now it seems that it's about the extent to which NFC is used (like CPU usage, network usage, etc.)

app/src/main/java/com/nervousfish/nervousfish/activities/NFCActivity.java
TheBonheurs: This is how it was done with bluetooth in verification activities though

app/src/main/java/com/nervousfish/nervousfish/modules/pairing/NFCHandler.java
TheBonheurs: This was on develop

app/src/main/java/com/nervousfish/nervousfish/activities/NFCActivity.java
jverbraeken: Well, that's than incorrectly done in bluetooth

app/src/main/java/com/nervousfish/nervousfish/activities/MainActivity.java
```diff
@@ -328,6 +338,46 @@ public void onClick(final SweetAlertDialog dialog) {
...
+                    .setContentText(description)
+                    .setCancelText(this.getString(R.string.no))
+                    .setConfirmText(this.getString(R.string.yes))
+                    .setConfirmClickListener(new SweetAlertDialog.OnSweetClickListener() {
+                        @Override
+                        public void onClick(final SweetAlertDialog dialog) {
+                            dialog.dismiss();
+
+                            LOGGER.info("Requesting to enable NFC");
+                            new SweetAlertDialog(MainActivity.this, SweetAlertDialog.NORMAL_TYPE)
```
ericcornelissen: I very much so dissaprove of this approach... what value does the second alert add? The first alert prompts the user with the message that they should enable NFC and the second alert prompts them with the request to do so...

Could you remove the second alert?

app/src/main/java/com/nervousfish/nervousfish/activities/NFCActivity.java
```diff
@@ -1,22 +1,143 @@
...
+     * {@inheritDoc}
+     */
+    @Override
+    public void onStop() {
+        super.onStop();
+
+        this.serviceLocator.unregisterFromEventBus(this);
+        LOGGER.info("Stopped NFCActivity");
+
+        super.onStop();
```
ericcornelissen: You don't have to call `super.onStop()` twice 😉

app/src/main/java/com/nervousfish/nervousfish/activities/VisualVerificationActivity.java
```diff
@@ -59,7 +59,7 @@ private void nextActivity() {
             LOGGER.info("Sending my profile with name: " + myProfile.getName() + ", public key: "
                     + myProfile.getPublicKey().toString());
             final Contact myProfileAsContact = new Contact(myProfile.getName(), new SimpleKey("simplekey", "73890ien"));
-            this.serviceLocator.getBluetoothHandler().send(myProfileAsContact);
+            this.serviceLocator.getBluetoothHandler().send(this.serviceLocator.getBluetoothHandler().objectToBytes(myProfileAsContact));
```
ericcornelissen: This seems very cumbersome to me... Now you always have to call `.objectToBytes()` when calling `.send()` on the BluetoothHandler, why not let the `.send()` method handle the `.objectToBytes()` part?

app/src/main/java/com/nervousfish/nervousfish/modules/pairing/NFCHandler.java
```diff
@@ -0,0 +1,71 @@
...
+             ObjectOutputStream oos = new ObjectOutputStream(bos)) {
+            oos.writeObject(new DataWrapper(object));
+            oos.flush();
+            bytes = bos.toByteArray();
+        }
+        return bytes;
+    }
+
+    @Override
+    public void send(final byte[] buffer) {
```
ericcornelissen: Missing JavaDoc on multiple occasions in this file

app/src/main/res/layout/activity_nfc.xml
```diff
@@ -1,5 +1,30 @@
...
+    <ImageView
+        tools:ignore = "UseCompoundDrawables"
+        android:id="@+id/nfc_gif"
+        android:layout_width="match_parent"
+        android:layout_height="match_parent"
+        android:contentDescription="@string/descr_gif"
+        android:src="@android:drawable/sym_def_app_icon" />
+
+    <TextView
+        tools:ignore = "UseCompoundDrawables"
```
ericcornelissen: Can you remove the spaces around the `=`? Also, do you need this ignore on every tag in this file?

app/src/main/res/values-nl/strings.xml
```diff
@@ -97,6 +97,13 @@
     <string name="contact_set_name">Geef contact een naam</string>
     <string name="contact_exists_message">Contact bestaat al, geef aub een unieke naam</string>
 
+    <!-- NFC -->
+    <string name="descr_gif">Dit zou een gif moeten zijn die nfc connection uitbeeldt.</string>
```
ericcornelissen: `<string name="nfc_image_description">Image visualizing how to use NFC</string>`

FYI: the `android:contentDescription` is intended for people that are blind or have bad eyesight 😉

TheBonheurs: @clenrock told me this should describe what the view was depicting so I think it does a good job on that 🙈

app/src/main/java/com/nervousfish/nervousfish/activities/VisualVerificationActivity.java
TheBonheurs: In a perfect world every serializable object knows how to serialize itself, as it is not the case we have to put this method somewhere like the pairing handlers. But according to @jverbraeken there had to be only one send with bytes while renaming send and only letting it return bytes so yeah pretty cumbersome but as intended by him...

app/src/main/java/com/nervousfish/nervousfish/modules/pairing/NFCHandler.java
TheBonheurs: This is fixed bu below this comment so git doesn't register it i guess

app/src/main/res/values-nl/strings.xml
ericcornelissen: Okay, it kinda does (I don't like the style tho, but thats another topic 🙂)

Could you still change the name tho 😇

app/src/main/java/com/nervousfish/nervousfish/activities/VisualVerificationActivity.java
ericcornelissen: > ...serializable object knows how to serialize itself...

How is that relevant, I don't see anything about serialization here...


> But according to @jverbraeken there had to be only one send with bytes while renaming send and only letting it return bytes so yeah pretty cumbersome but as intended by him

Not sure what you mean there, but why not have two `send()` methods that accept parameters of different types?

app/src/main/res/layout/activity_nfc.xml
TheBonheurs: Didn't work the first two tries so I added them above but forgot to remove lol

app/src/main/java/com/nervousfish/nervousfish/activities/VisualVerificationActivity.java
TheBonheurs: objectToBytes = serialization. What you mean is go back to the old version with two sends but free the send with an object from the serialization task and put it in a third method named objectToBytes? Sounds good but Joost definitely convince me to do it this way I just can't remember the specifics, it had to do with the differences between the handlers or something like that.

ericcornelissen: @jverbraeken, can you help here?

jverbraeken: > In a perfect world every serializable object knows how to serialize itself
There is a reason why I want that every class has a readObject and a writeObject or a SerializationProxy...

The reason why I promoted the method objectToBytes is because
- we do not want duplicate code, although both NFC and Bluetooth have to convert a message to bytes
- we want a function to have a single task, and send(Serializable) both converts the serializable object to bytes and makes sure that these bytes are sent
- I hate it as much as you do, but it seems that it's not possible to the NFCHandler send messages. It has to be done by the current activity. You cannot "send" messages, but you fill a buffer when the activity starts with the data you want to beam to the other activity. Android beams the buffer constantly to the ether. The send() method is therefore completely useless for NFC.

ericcornelissen: > we do not want duplicate code, although both NFC and Bluetooth have to convert a message to bytes

Not sure if you are aware of this, but you're actually introducing duplicate code because now you constantly have to convert these things to bytes separately. I don't know where the other duplicate code would appear since the problem you're describing is based in a single class. Can you explain where we would have had duplicate code if it where not for this change?


> we want a function to have a single task, and send(Serializable) both converts the serializable object to bytes and makes sure that these bytes are sent

Indeed a single task, **send data**. I don't care what data, is it a string? send it. Is it an object? send it. I don't care how it has to be send... Ow it has to be send as a byte-array? Well let the `send()` method handle that.


> I hate it as much as you do, but it seems that it's not possible to the NFCHandler send messages. It has to be done by the current activity. You cannot "send" messages, but you fill a buffer when the activity starts with the data you want to beam to the other activity. Android beams the buffer constantly to the ether. The send() method is therefore completely useless for NFC.

OK, sure. So we need a send method that accepts byte-arrays and one that accepts object? That works for me...

TheBonheurs:
Should've fixed all the remarks @jverbraeken  @ericcornelissen , tell me if I missed something

app/src/main/java/com/nervousfish/nervousfish/activities/VisualVerificationActivity.java
jverbraeken: > Not sure if you are aware of this, but you're actually introducing duplicate code because now you constantly have to convert these things to bytes separately. I don't know where the other duplicate code would appear since the problem you're describing is based in a single class. Can you explain where we would have had duplicate code if it where not for this change?
Yes, I can. NFC needs the message as bytes in order to send it. However, it cannot use the send() method of APairingHandler, because it has to send the bytes in the activity itself. Because with the current implementation it is impossible to convert an object to bytes without duplicating the code in the send() function of APairingHandler, we want to expose that functionality in APairingHandler to classes like NFC.

> Indeed a single task, send data. I don't care what data, is it a string? send it. Is it an object? send it. I don't care how it has to be send... Ow it has to be send as a byte-array? Well let the send() method handle that.

You could look at it that way, but **from the view of NFCActivity** the method does convert the object to bytes (great, love that, I need that functinality!) AND the method sends the bytes using the send method implemented by Bluetooth, NFC or QRHandler (Uh oh, I don't want that! It's Android so NFCHandler cannot do that at all, I have to do that in the NFC activity myself)

> OK, sure. So we need a send method that accepts byte-arrays and one that accepts object? That works for me...

Sadly, I thought about that as well, but it is not possible it seems. The send method doesn't return anything, but the method createNdefMessage (which MUST be implemented and is called by Android) returns and NdefMessage and send doesn't return such a message of course. Bluetooth doesn't need that, QR doesn't need an NdefMessage, only NFC. Therefore we cannot delegate the functionality of createNdefMessage to NFCHandler and therefore we cannot use NFCHandler at all.

jverbraeken:
@ericcornelissen You can find under VisualVerificationActivity.java (Outdated) under the large image provided by Stas my response

app/src/main/java/com/nervousfish/nervousfish/activities/VisualVerificationActivity.java
ericcornelissen: > Yes, I can. NFC needs the message as bytes in order to send it. However, it cannot use the send() method of APairingHandler

I was not aware of that. So OK, the `objectToBytes()` methods must be exposed, but you can still have two `send()` methods (or even one, not sure if we really need to) where one of them accepts any object and uses the `objectToBytes()` methods to get the bytes of that object

jverbraeken: One send method, namely send(byte[]) should be enough indeed.

jverbraeken: I don't think it's a problem if we have two methods, but I don't have a problem with one method either.
@ericcornelissen @TheBonheurs Do you think we should remove send(Serializable) because it's redundant although it simplifies the sending process by a single call?

app/src/main/java/com/nervousfish/nervousfish/modules/pairing/NFCHandler.java
```diff
@@ -0,0 +1,76 @@
...
+     */
+    public static ModuleWrapper<NFCHandler> newInstance(final IServiceLocator serviceLocator) {
+        return new ModuleWrapper<>(new NFCHandler(serviceLocator));
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public byte[] objectToBytes(final Serializable object) throws IOException {
```
jverbraeken: This method can be removed as it is already defined in APairingHandler

app/src/main/java/com/nervousfish/nervousfish/activities/VisualVerificationActivity.java
TheBonheurs: If you look at my latest changes I went from 1 to 2 methods, actually this was there since @ericcornelissen posted his 2nd comment :p

ericcornelissen: > One send method, namely send(byte[]) should be enough indeed.

😕 Thats not what I meant... by "or even one, not sure if we really need two" I meant that I'm not sure if we still needed the `send(byte[])` version because NFC just needs the bytes and doesn't send anything through the handler.


> I don't think it's a problem if we have two methods, but I don't have a problem with one method either.

I do have a problem with it, that is why I started this thread 😅


> If you look at my latest changes I went from 1 to 2 methods, actually this was there since @ericcornelissen posted his 2nd comment :p

I would actually like to see just the other method, so `send(Object)` ([looking at the diff](https://github.com/ericcornelissen/NervousFish/pull/231/files/2058483bfce4465c711e98be6fb407e4655b5248#diff-12207c280a7c5d460c92841b36106522R62) you're now using `send(byte[])`, right?)

app/src/main/java/com/nervousfish/nervousfish/activities/BluetoothConnectionActivity.java
```diff
@@ -54,7 +54,6 @@
     // Used to fill the listview of newly discovered Bluetooth devices
     private ArrayAdapter<String> newDevicesArrayAdapter;
     // Used to fill the listview of paired Bluetooth devices
-    private ArrayAdapter<String> pairedDevicesArrayAdapter;
```
clenrock: Can you remove the unused comment "// Used to fill the listview of paired Bluetooth devices"

TheBonheurs: Sure

app/src/main/java/com/nervousfish/nervousfish/activities/VisualVerificationActivity.java
TheBonheurs: Hmm that's weird. I'm pretty sure I had changed it...

TheBonheurs: joost
And this can't be committed as git says it's already on the branch

jverbraeken: @ericcornelissen I would certainly like to have it with a single send method, but I still don't understand how that's possible. NFCActivity must implement createNdefMessage. createNdefMessage must return an NdefMessage. No way around those requirements. The send method is void - cannot return an NdefMessage - so how do you propose to call send in createNdefMessage?

ericcornelissen: > Hmm that's weird. I'm pretty sure I had changed it...

You're right, I like it this way 👍 


> I would certainly like to have it with a single send method, but I still don't understand how that's possible...

Not sure what you mean, it seems @TheBonheurs has been able to do it 😕

PR #238, opened 2017-06-09
Style errors

jverbraeken:
- Relevant Issues: -
- Related Pull Requests: -

* * *

## What
This Pull Request adds to the repository a maximum consistent code style by adding all missing "this" keywords, making all classes final that should be final, etc. This is done for most part automatically by Intellij Idea.

## Why
This Pull Request is needed because the code style was not consistent everywhere (e.g. the usage of 'this')

## How
This feature can be viewed/tested within the project by static code analysis

## Alternative implementation


app/src/main/java/com/nervousfish/nervousfish/activities/WaitActivity.java
```diff
@@ -33,6 +34,7 @@
 
     private static final Logger LOGGER = LoggerFactory.getLogger("WaitActivity");
     private IServiceLocator serviceLocator;
+    private IDatabase database;
```
ericcornelissen: Not sure if you need this attribute since you're only using it in the `onNewDataReceivedEvent()` method 🤔

app/src/main/java/com/nervousfish/nervousfish/modules/constants/Constants.java
```diff
@@ -37,6 +39,10 @@
      */
     private Constants(final IServiceLocator serviceLocator) {
         this.androidFilesDir = serviceLocator.getAndroidFilesDir();
+        //noinspection StringConcatenationMissingWhitespace because this is a file path
+        this.databaseContactsPath = this.androidFilesDir + Constants.DB_CONTACTS_PATH;
+        //noinspection StringConcatenationMissingWhitespace because this is a file path
+        this.databaseUserPath = this.androidFilesDir + Constants.DB_USERDATA_PATH;
```
ericcornelissen: I'm against this change, especially since you're now also writing this concatenation in the `readObject()` method (see [here](https://github.com/ericcornelissen/NervousFish/blob/2a4aabfecd1c3391936a0ce27a8dece5091ccd14/app/src/main/java/com/nervousfish/nervousfish/modules/constants/Constants.java#L101))

app/src/main/java/com/nervousfish/nervousfish/modules/constants/Constants.java
```diff
@@ -37,6 +39,10 @@
      */
     private Constants(final IServiceLocator serviceLocator) {
         this.androidFilesDir = serviceLocator.getAndroidFilesDir();
+        //noinspection StringConcatenationMissingWhitespace because this is a file path
+        this.databaseContactsPath = this.androidFilesDir + Constants.DB_CONTACTS_PATH;
+        //noinspection StringConcatenationMissingWhitespace because this is a file path
```
ericcornelissen: Where are these comments coming from?

app/src/main/java/com/nervousfish/nervousfish/modules/database/GsonDatabaseAdapter.java
```diff
@@ -314,31 +315,31 @@ private void initializeContacts() throws IOException {
...
-     * @throws InvalidObjectException Thrown when the state of the class is unstable
      */
-    private void ensureClassInvariant() throws InvalidObjectException {
-        assertNotNull(this.contactsPath);
-        assertNotNull(this.profilesPath);
-        assertNotNull(this.fileSystem);
+    private void ensureClassInvariant() {
+        assertThat(this.contactsPath, notNullValue());
+        assertThat(this.profilesPath, notNullValue());
+        assertThat(this.fileSystem, notNullValue());
```
ericcornelissen: What is wrong with `assertNotNull()`?

app/src/main/java/com/nervousfish/nervousfish/modules/database/GsonKeyAdapter.java
```diff
@@ -18,50 +18,43 @@
...
-        super();
-    }
-
-    /**
      * {@inheritDoc}
      */
     @Override
-    public void write(final JsonWriter writer, final IKey key) throws IOException {
-        writer.beginArray();
+    public void write(final JsonWriter jsonWriter, final IKey t) throws IOException {
```
ericcornelissen: How is `t` a better name for that parameter?

app/src/main/java/com/nervousfish/nervousfish/activities/CreateProfileHelper.java
```diff
@@ -11,7 +11,7 @@
 /**
  * Helper method for the logical functionality of the {@link CreateProfileActivity}.
  */
-class CreateProfileHelper {
+final class CreateProfileHelper {
```
TheBonheurs: Does this have to be final? What if we want to extend thi class in the future?

app/src/main/java/com/nervousfish/nervousfish/activities/CreateProfileHelper.java
```diff
@@ -56,7 +56,7 @@ KeyPair generateKeyPair(final IKey.Types keyType) {
      */
     boolean validateName(final EditText input) {
         final String name = input.getText().toString();
-        if (this.isValidName(name)) {
```
TheBonheurs: Why are you removing the this references here when previously you were adding them?

app/src/main/java/com/nervousfish/nervousfish/activities/CreateProfileHelper.java
```diff
@@ -119,7 +119,7 @@ private boolean isValidName(final String name) {
      * @param password The password that has been entered.
      * @return a {@link boolean} indicating whether or not the password is valid.
      */
-    private boolean isValidPassword(final String password) {
+    private static boolean isValidPassword(final String password) {
```
TheBonheurs: Why are these two methods static and not the validateName/Password methods?

app/src/main/java/com/nervousfish/nervousfish/activities/MainActivity.java
```diff
@@ -5,18 +5,19 @@
 import android.content.ComponentName;
 import android.content.Intent;
 import android.os.Bundle;
-import com.github.clans.fab.FloatingActionButton;
 import android.support.v7.app.AppCompatActivity;
 import android.support.v7.widget.Toolbar;
 import android.view.View;
 
+import com.github.clans.fab.FloatingActionButton;
```
TheBonheurs: Nice one 🥇

app/src/main/java/com/nervousfish/nervousfish/activities/MainActivity.java
```diff
@@ -290,15 +292,15 @@ private void enableBluetooth(final boolean buttonClicked) {
                     .setConfirmText(this.getString(R.string.yes))
                     .setConfirmClickListener(new SweetAlertDialog.OnSweetClickListener() {
                         @Override
-                        public void onClick(final SweetAlertDialog dialog) {
-                            dialog.dismiss();
+                        public void onClick(final SweetAlertDialog sweetAlertDialog) {
```
TheBonheurs: Overkill imo but don't mind it

app/src/main/java/com/nervousfish/nervousfish/activities/MainActivity.java
```diff
@@ -276,12 +280,10 @@ private void enableBluetooth(final boolean buttonClicked) {
...
-            final String description;
-            if (buttonClicked) {
-                description = this.getString(R.string.popup_enable_bluetooth_exchange);
-            } else {
-                description = this.getString(R.string.popup_enable_bluetooth_findable);
-            }
+            final String description = this.getString(
+                    buttonClicked
+                            ? R.string.popup_enable_bluetooth_exchange
+                            : R.string.popup_enable_bluetooth_findable);
```
TheBonheurs: This formatting looks awful to me 😆 , not the ternary operator but the newlines btw

app/src/main/java/com/nervousfish/nervousfish/activities/QRExchangeKeyActivity.java
```diff
@@ -174,8 +176,9 @@ public void onClick(final DialogInterface dialog, final int which) {
         }
     }
 
-    private final class EditNameClickListener implements DialogInterface.OnClickListener {
-
+    private static final class EditNameClickListener implements DialogInterface.OnClickListener {
+        private final Activity activity;
+        private final IDatabase database;
```
TheBonheurs: Local var?

app/src/main/java/com/nervousfish/nervousfish/modules/database/GsonDatabaseAdapter.java
```diff
@@ -300,10 +300,11 @@ private void initializeContacts() throws IOException {
...
             LOGGER.warn("Contacts part of the database didn't exist");
-            final Writer writer = this.fileSystem.getWriter(this.contactsPath);
-            writer.write("[]");
-            writer.close();
-            LOGGER.info("Created the contacts part of the database");
+            try (Writer writer = this.fileSystem.getWriter(this.contactsPath)) {
+                writer.write("[]");
+                writer.close();
+                LOGGER.info("Created the contacts part of the database");
+            }
```
TheBonheurs: no catch?

app/src/main/java/com/nervousfish/nervousfish/modules/database/GsonDatabaseAdapter.java
```diff
@@ -314,31 +315,31 @@ private void initializeContacts() throws IOException {
...
-            final Writer writer = this.fileSystem.getWriter(this.profilesPath);
-            writer.write("[]");
-            writer.close();
-            LOGGER.info("Created the part of the database: %s", this.profilesPath);
+            LOGGER.warn("Part of the database didn't exist: {}", this.profilesPath);
+            try (Writer writer = this.fileSystem.getWriter(this.profilesPath)) {
+                writer.write("[]");
+                writer.close();
+                LOGGER.info("Created the part of the database: {}", this.profilesPath);
+            }
```
TheBonheurs: Same as above

app/src/main/java/com/nervousfish/nervousfish/modules/pairing/APairingHandler.java
```diff
@@ -37,15 +38,15 @@
      * {@inheritDoc}
      */
     @Override
+    @DesignedForExtension
```
TheBonheurs: What does this do, just curious? 😲

app/src/main/java/com/nervousfish/nervousfish/modules/pairing/DataWrapper.java
```diff
@@ -62,7 +62,7 @@ private void readObject(final ObjectInputStream stream) throws InvalidObjectExce
      * ordinary serialization - is far more dangerous
      */
     // A private constructor is safer than a hidden constructor
-    @SuppressWarnings("PMD.AccessorClassGeneration")
+    @SuppressWarnings({"PMD.AccessorClassGeneration", "SerializableHasSerializationMethods"})
```
TheBonheurs: Comment on the suppression

app/src/main/java/com/nervousfish/nervousfish/activities/BluetoothConnectionActivity.java
```diff
@@ -50,7 +50,7 @@
     private BluetoothAdapter bluetoothAdapter;
     private IBluetoothHandler bluetoothHandler;
     private Set<BluetoothDevice> pairedDevices;
-    private boolean isMaster = false;
+    private boolean isMaster;
```
jverbraeken: Technically speaking nothing, but if you set Intellij checking on the highest level, then it gives a warning that this clutters the codebase. For experienced developers it's clear that the value is initialized to false, so the assignment is redundant

app/src/main/java/com/nervousfish/nervousfish/activities/BluetoothConnectionActivity.java
```diff
@@ -104,6 +105,10 @@ protected void onCreate(final Bundle savedInstanceState) {
         final IntentFilter filter = new IntentFilter(BluetoothAdapter.ACTION_DISCOVERY_FINISHED);
         this.registerReceiver(this.broadcastReceiver, filter);
 
+        this.serviceLocator.registerToEventBus(this);
```
jverbraeken: Oops, shouldn't have done that

app/src/main/java/com/nervousfish/nervousfish/activities/MainActivity.java
```diff
@@ -49,10 +51,11 @@
 public final class MainActivity extends AppCompatActivity {
 
     private static final Logger LOGGER = LoggerFactory.getLogger("MainActivity");
-    private static final int REQUEST_CODE_ENABLE_BLUETOOTH_ON_START = 100;
-    private static final int REQUEST_CODE_ENABLE_BLUETOOTH_ON_BUTTON_CLICK = 200;
+    private static final int ENABLE_BLUETOOTH_ON_START = 100;
+    private static final int ENABLE_BLUETOOTH_ON_BUTTON_CLICK = 200;
```
jverbraeken: They were too long

app/src/main/java/com/nervousfish/nervousfish/activities/MainActivity.java
```diff
@@ -49,10 +51,11 @@
...
 
     private IServiceLocator serviceLocator;
+    private IDatabase database;
```
jverbraeken: Also in onResume

app/src/main/java/com/nervousfish/nervousfish/activities/MainActivity.java
```diff
@@ -69,26 +72,28 @@ protected void onCreate(final Bundle savedInstanceState) {
...
-        if (getSupportActionBar() != null) {
-            getSupportActionBar().setDisplayShowTitleEnabled(false);
+        if (this.getSupportActionBar() != null) {
+            this.getSupportActionBar().setDisplayShowTitleEnabled(false);
         }
 
         // Fill database with demo data
         try {
-            this.contacts = this.serviceLocator.getDatabase().getAllContacts();
+            this.contacts = this.database.getAllContacts();
```
jverbraeken: That was a bug, good catch!

app/src/main/java/com/nervousfish/nervousfish/activities/MainActivity.java
```diff
@@ -69,26 +72,28 @@ protected void onCreate(final Bundle savedInstanceState) {
...
             LOGGER.error("Failed to retrieve contacts from database", e);
         }
 
+        final IBluetoothHandler bluetoothHandler = this.serviceLocator.getBluetoothHandler();
         // Start Bluetooth
         try {
-            this.serviceLocator.getBluetoothHandler().start();
-        } catch (NoBluetoothException e) {
-            LOGGER.info("Bluetooth not available on device, disabling button");
+            //noinspection LawOfDemeter because we don't want to clutter the service locator
```
jverbraeken: Suppression of Intellij warning

app/src/main/java/com/nervousfish/nervousfish/activities/MainActivity.java
```diff
@@ -122,8 +127,7 @@ protected void onResume() {
         super.onResume();
 
         try {
-            final IDatabase database = this.serviceLocator.getDatabase();
-            this.contacts = database.getAllContacts();
+            this.contacts = this.database.getAllContacts();
```
jverbraeken: Fixed

app/src/main/java/com/nervousfish/nervousfish/activities/QRExchangeKeyActivity.java
```diff
@@ -184,7 +187,9 @@ public void onClick(final DialogInterface dialog, final int which) {
          * @param editName The textinput from which we show and get the name from.
          * @param key The key made from the QR code.
          */
-        private EditNameClickListener(final EditText editName, final IKey key) {
+        private EditNameClickListener(final Activity activity, final IServiceLocator serviceLocator, final EditText editName, final IKey key) {
```
jverbraeken: If I integrate it in the activity, it becomes a too-large anonymous class. In my opinion it's better to dependency inject stuff than to ask for it, see https://softwareengineering.stackexchange.com/questions/157526/explanation-on-how-tell-dont-ask-is-considered-good-oo

app/src/main/java/com/nervousfish/nervousfish/activities/WaitActivity.java
jverbraeken: To follow the law of Demeter, that is, you should only communicate with your friends, not with friends of friends. I think that the law of Demeter really makes code a lot less error-prone and reduces the change on null-pointers exponentially. A class is a friend if it's in a field of the class.

app/src/main/java/com/nervousfish/nervousfish/modules/constants/Constants.java
jverbraeken: To suppress Intellij warnings, I'm thinking about removing them, they're not very aesthetic

app/src/main/java/com/nervousfish/nervousfish/modules/constants/Constants.java
jverbraeken: I did it, because it's inefficient to concatentate Strings every single time you call the getter. You only want to do that once.

jverbraeken: However, I agree that code duplication is bad, so I've made a helper method to solve that

app/src/main/java/com/nervousfish/nervousfish/modules/database/GsonDatabaseAdapter.java
jverbraeken: Hamcrest is more readable.

app/src/main/java/com/nervousfish/nervousfish/modules/database/GsonKeyAdapter.java
jverbraeken: Because the parameter name of the interface is called t, see the javadoc, and it can be confusing if the parameter name of the superclass (the interface) is different from the parameter name of the implementation

app/src/main/java/com/nervousfish/nervousfish/activities/CreateProfileHelper.java
jverbraeken: We've no reason to not make it final. If we want to extend the class in the future, then you we can always use a decorator (good coding style anyway). Otherwise we'll have to eliminate the class's self-use of overridable methods entirely and well... That costs a lot of time. Additionally, when you decide to make a class extendable, you should actually also test it, which costs even more time.

app/src/main/java/com/nervousfish/nervousfish/activities/CreateProfileHelper.java
jverbraeken: Because isValidName doesn't belong to this, but to CreateProfileHelper. The method is static

app/src/main/java/com/nervousfish/nervousfish/activities/CreateProfileHelper.java
jverbraeken: Because validateName/password use member fields and isValidPassword not. Actually, I didn't change this myself, that would cost too much time to do in the last few weeks xD . Just clicked auto cleanup in Intellij and I liked the result

app/src/main/java/com/nervousfish/nervousfish/activities/MainActivity.java
jverbraeken: Ctrl-alt-shift-L is absolutely amazing, I agree!

app/src/main/java/com/nervousfish/nervousfish/activities/MainActivity.java
jverbraeken: Yeah, but otherwise the line is a little bit too long

app/src/main/java/com/nervousfish/nervousfish/activities/QRExchangeKeyActivity.java
jverbraeken: Yes, local var

app/src/main/java/com/nervousfish/nervousfish/modules/database/GsonDatabaseAdapter.java
jverbraeken: No, I'm only using the try to make sure that the resources I allocate are released, not to catch errors

app/src/main/java/com/nervousfish/nervousfish/modules/database/GsonDatabaseAdapter.java
jverbraeken: Same as above

app/src/main/java/com/nervousfish/nervousfish/modules/pairing/APairingHandler.java
jverbraeken: So that FindBugs knows that this method is designed for extension and won't complain because it isn't final

app/src/main/java/com/nervousfish/nervousfish/activities/BluetoothConnectionActivity.java
ericcornelissen: I think its is always better to explicitly set this value, even the most experienced programmer can make an error and interpret this wrongly...

app/src/main/java/com/nervousfish/nervousfish/activities/WaitActivity.java
ericcornelissen: Okey, but I hope you're aware the "preventing null pointer errors" doesn't hold here since you can't make the variable final (because it must be set in onCreate)

app/src/main/java/com/nervousfish/nervousfish/modules/database/GsonDatabaseAdapter.java
ericcornelissen: Is assertNotNull not readable ?

app/src/main/java/com/nervousfish/nervousfish/activities/BluetoothConnectionActivity.java
jverbraeken: Hmm, it seems that Intellij has no option to give a warning when you forget to initialize a field. In my opinion it's unimportant if we do or do not initialize fields to their default values. Consistency is more important I think and because Intellij can only (I agree that that should not be the case) give a warning if you have a redundant initializer, we can only enforce consistent code if we do not initialize fields to their default value.

app/src/main/java/com/nervousfish/nervousfish/activities/QRExchangeKeyActivity.java
TheBonheurs: This isn't one though xd

app/src/main/java/com/nervousfish/nervousfish/modules/database/GsonDatabaseAdapter.java
jverbraeken: Well, ok, in this case it doesn't really matter. Just a personal preference.
However, there are a lot more reasons to prefer assertThat over the old JUnit functions, see https://objectpartners.com/2013/09/18/the-benefits-of-using-assertthat-over-other-assert-methods-in-unit-tests/

app/src/main/java/com/nervousfish/nervousfish/activities/QRExchangeKeyActivity.java
jverbraeken: That's true, but we cannot make it a local variable, because the constructor needs the key, and we don't know that key in onCreate yet

app/src/main/java/com/nervousfish/nervousfish/activities/QRExchangeKeyActivity.java
ericcornelissen: Ok sure, but how about the `ServiceLocator` ?

app/src/main/java/com/nervousfish/nervousfish/activities/MainActivity.java
ericcornelissen: Maybe:
```java
buttonClicked ?
  R.string.popup_enable_bluetooth_exchange : R.string.popup_enable_bluetooth_findable
```

or
```java
buttonClicked ? R.string.popup_enable_bluetooth_exchange :
  R.string.popup_enable_bluetooth_findable
```

app/src/main/java/com/nervousfish/nervousfish/modules/database/GsonKeyAdapter.java
ericcornelissen: 🤔 not sure if I agree when that is the name used in the interface...

app/src/main/java/com/nervousfish/nervousfish/activities/MainActivity.java
ericcornelissen: 👍

app/src/main/java/com/nervousfish/nervousfish/modules/database/GsonDatabaseAdapter.java
ericcornelissen: 👍

app/src/main/java/com/nervousfish/nervousfish/activities/MainActivity.java
ericcornelissen: Missed that 😅

app/src/main/java/com/nervousfish/nervousfish/activities/BluetoothConnectionActivity.java
ericcornelissen: I'm not sure if enforcing code style through an IDE is good practice either, no one is forced to use IntelliJ/Android studio... Preferably we could enforce this through Checkstyle or some kind of linter 🤔

jverbraeken: Checkstyle can only enforce, just like Intellij, only that variables are not initiallized to their default values, see http://checkstyle.sourceforge.net/apidocs/com/puppycrawl/tools/checkstyle/checks/coding/ExplicitInitializationCheck.html

jverbraeken: I can apply that rule if you want

app/src/main/java/com/nervousfish/nervousfish/activities/QRExchangeKeyActivity.java
jverbraeken: Nice one, fixed it!

app/src/main/java/com/nervousfish/nervousfish/activities/BluetoothConnectionActivity.java
ericcornelissen: Thats okay

app/src/main/java/com/nervousfish/nervousfish/activities/WaitActivity.java
jverbraeken: Okay. However, I still don't think that following the Law of Demeter is bad practice to follow, see http://www.virtuouscode.com/2011/07/05/demeter-its-not-just-a-good-idea-its-the-law/.

app/src/main/java/com/nervousfish/nervousfish/modules/database/GsonKeyAdapter.java
jverbraeken: Ok, changed it back and suppressed the warning

app/src/main/java/com/nervousfish/nervousfish/activities/WaitActivity.java
ericcornelissen: I'm OK with the change 😄

app/src/main/java/com/nervousfish/nervousfish/activities/MainActivity.java
ericcornelissen: Then what do you mean by "because we don't want to clutter the service locator"?

jverbraeken: Because I would have to add a method named "startBluetoothHandler" in ServiceLocator and well.. That's not what the service locator should have imo

ericcornelissen: OK, but why then is `bluetoothHandler` not an attribute of `MainActivity`?

app/src/main/java/com/nervousfish/nervousfish/modules/constants/Constants.java
ericcornelissen: 👍

PR #242, opened 2017-06-12
Fix Bluetooth cancel methods

clenrock:
- Relevant Issues: #197
- Related Pull Requests: #195 #208

* * *

## What
This Pull Request adds to the repository a fix for the cancel methods in Bluetooth

## Why
This Pull Request is needed because the socket of connectthread is never closed when the user is never connected, but only initiates the request.

## How
This feature can be viewed/tested within the project by verifying that Bluetooth still works.

## Alternative implementation

## Notes

~~#195 should be merged first~~


ericcornelissen:
Merged in b961218

PR #248, opened 2017-06-12
Let pairing button be above listview

jverbraeken:
- Relevant Issues: -
- Related Pull Requests: -

* * *

## What
This Pull Request adds to the repository the placement of the pairing button above the listview in MainActivity

## Why
This Pull Request is needed because you couldn't click on pairing buttons when the listview overlapped the pairing buttons

## How
This feature can be viewed/tested within the project by adding a lot of contacts and verifying that the button is above the listview elements

## Alternative implementation

## Notes

The sorting button is now wrapped in a LinearLayout because RelativeLayout doesn't support margins (see https://stackoverflow.com/questions/40041415/android-relativelayout-marginbottom-not-working)


app/src/main/java/com/nervousfish/nervousfish/activities/MainActivity.java
```diff
@@ -355,6 +356,22 @@ public void onClick(final SweetAlertDialog dialog) {
     }
 
     List<Contact> getContacts() {
+        contacts.add(new Contact("foo", new SimpleKey("bar", "baz")));
```
ericcornelissen: Interesting change 🤔 this shouldn't be part of the PR...

PR #252, opened 2017-06-13
Encrypt the database, closes #202 and closes #44

### LINKED ISSUE #44: Remove functionality that allows you to pass the login screen when no password is entered
- Parent issues: -
- Relevant issues: -
- Estimated effort: -

* * *

## User story
As a user I don't want that people are able to login into the application without providing a password so that my data is safe.

## What
Currently, the LoginActivity has some custom logic that allows you to login without providing a password (for development purposes). This should be removed or replaced with a proper way to enable it in development mode only.

## Why
This issue is needed for the project because people shouldn't be allowed to access the application without a password if the user wants so.

## How
This feature can be added by removing the functionality that allows you to pass the login screen when no password is entered

## Definition of done

- 

## Notes
⚠️ THIS ISSUE DOES NOT REQUIRE TOO MUCH ATTENTION RIGHT NOW ⚠️ 

The code in question can be found [here](https://github.com/ericcornelissen/NervousFish/blob/96acc8e91c2b4a817f4d83a09a84eecddba4b7e8/app/src/main/java/com/nervousfish/nervousfish/LoginActivity.java#L46)

### LINKED ISSUE #202: Single encrypted database

KilianCallebaut:
- Relevant Issues: #4 #202 #44
- Related Pull Requests: -

* * *

## What
Added functionality for checking passwords, encrypting strings using various methods and storing the database encrypted.

## Why
This Pull Request is needed because it adds more security to the user's profile and data.

## How
This feature can be viewed/tested within the project by checking if the database is stored encrypted after each operation. Also, one could see if they can log in using their created password and no other.

## Alternative implementation

## Notes
There are a lot of tests missing, since there kept popping up problems and other collaborators needed parts of the system. These will be added later. Also in the encrypted saver there are methods commented out. This is because of circumstances made it so that these were no longer necessairy encryption methods, yet since they are working implementations other people could perhaps use them.


jverbraeken:
Can you make several smaller PR's instead of a single very large one next time? :P

app/build.gradle
```diff
@@ -64,6 +67,11 @@ dependencies {
 
     testCompile 'junit:junit:4.12'
     testCompile 'org.mockito:mockito-core:2.8.9'
+    testCompile 'org.powermock:powermock-api-mockito:1.6.1'
```
jverbraeken: Can you remove this? These dependencies aren't used anywhere

app/src/main/java/com/nervousfish/nervousfish/data_objects/Contact.java
```diff
@@ -59,8 +59,14 @@ public boolean equals(final Object o) {
         }
 
         final Contact that = (Contact) o;
-        return this.name.equals(that.name)
-                && this.keys.equals(that.keys);
+        for (final IKey key: this.keys) {
```
jverbraeken: The old way was good, see https://stackoverflow.com/questions/1075656/simple-way-to-find-if-two-different-lists-contain-exactly-the-same-elements

app/src/main/java/com/nervousfish/nervousfish/data_objects/Database.java
```diff
@@ -0,0 +1,107 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * A database object to write and encrypt to file
+ */
+public final class Database implements Serializable {
+    private static final long serialVersionUID = -4715364587956219157L;
+    private final List<Contact> contacts;
+    private Profile profile;
```
jverbraeken: Profile MUST be Serializable, otherwise you can get a very nasty bug causing Database not to be Serializable

app/src/main/java/com/nervousfish/nervousfish/data_objects/Database.java
```diff
@@ -0,0 +1,107 @@
...
+     * Sets the user's profile to a new profile
+     * @param profile The new profile.
+     */
+    public void setProfile(final Profile profile) { this.profile = profile; }
+
+    /**
+     * Serialize the created proxy instead of this instance.
+     */
+    private Object writeReplace() {
+        return new SerializationProxy(this);
```
jverbraeken: Should be Database.SerializationProxy, because that's the code style that's mostly used in this project (apparently). Press Ctrl-Alt-S -> Editor -> Inspections -> Manage -> Import. https://www.dropbox.com/s/wryiggq2x9yr0tz/Project_Default.xml?dl=0
Hover over SerializationProxy and press Cleanup Code. That will automatically fix all "style errors" in the whole document of that type

app/src/main/java/com/nervousfish/nervousfish/data_objects/Database.java
```diff
@@ -0,0 +1,107 @@
...
+            this.profile = database.profile;
+        }
+
+        /**
+         * Not to be called by the user - resolves a new object of this proxy
+         * @return The object resolved by this proxy
+         */
+        private Object readResolve() {
+            final List<Contact> contactsList = new ArrayList<>();
+            for (final Contact contact : contacts) {
```
jverbraeken: Collections.addAll(contactsList, this.contacts)

app/src/main/java/com/nervousfish/nervousfish/data_objects/DatabasePass.java
```diff
@@ -0,0 +1,76 @@
...
+import com.nervousfish.nervousfish.ConstantKeywords;
+import com.nervousfish.nervousfish.modules.cryptography.EncryptedSaver;
+
+import java.io.InvalidObjectException;
+import java.io.ObjectInputStream;
+import java.io.Serializable;
+
+
+/**
+ * A databasePass object to save the enccryption elements
```
jverbraeken: Can you elaborate a bit more on this? I have no idea what this class is for. Maybe you should also consider renaming the class, because we don't use abbreviations anywhere in the project (assuming that Pass is the abbreviation of something)

app/src/main/java/com/nervousfish/nervousfish/data_objects/Profile.java
```diff
@@ -1,57 +1,69 @@
...
+import com.nervousfish.nervousfish.ConstantKeywords;
+
+import java.io.InvalidObjectException;
+import java.io.ObjectInputStream;
+import java.io.Serializable;
+import java.util.Arrays;
+import java.util.List;
+
 /**
  * An Profile POJO to store a name, public key and private key.
```
jverbraeken: This javadoc is outdated

app/src/main/java/com/nervousfish/nervousfish/data_objects/Profile.java
```diff
@@ -1,57 +1,69 @@
...
-     * Get the private {@link IKey} of the {@link Profile}.
-     *
-     * @return The private {@link IKey}.
+     * Returns the (first) public key of the profile.
+     * @return the (first) public key.
      */
-    public IKey getPrivateKey() {
-        return this.keyPair.getPrivateKey();
-    }
+    public IKey getPublicKey() { return keyPairs.get(0).getPublicKey(); }
```
jverbraeken: Can you add a TODO here that we should let the user choose which keypair is currently "active"

app/src/main/java/com/nervousfish/nervousfish/data_objects/Profile.java
```diff
@@ -63,16 +75,67 @@ public boolean equals(final Object o) {
         }
 
         final Profile that = (Profile) o;
-        return this.name.equals(that.name)
-            && this.keyPair.equals(that.keyPair);
+        if (this.keyPairs.size() != that.keyPairs.size()) {
+            return false;
+        }
+        for (final KeyPair kp : this.keyPairs) {
```
jverbraeken: The old one was better.

app/src/main/java/com/nervousfish/nervousfish/modules/cryptography/EncryptedSaver.java
```diff
@@ -0,0 +1,266 @@
...
+
+    }
+
+    /**
+     * Generates a salt bytearray based on a seed and with a given length.
+     * @param seed  The seed to keep the salt consistent.
+     * @param length The length the salt should be.
+     * @return  The salt bytearray.
+     */
+    /*
```
jverbraeken: Commented code

app/src/main/java/com/nervousfish/nervousfish/modules/cryptography/EncryptedSaver.java
```diff
@@ -0,0 +1,266 @@
...
+     */
+    public static String hashWithoutSalt(final String pass) {
+        try {
+            final MessageDigest digest = MessageDigest.getInstance("SHA-256");
+            digest.reset();
+            final byte[] hash = digest.digest(pass.getBytes(StandardCharsets.UTF_8));
+            LOGGER.info("Hashed the pass with SHA-256, no salt");
+            return new String(hash, UTF_8);
+        } catch (NoSuchAlgorithmException e) {
+            LOGGER.error("SHA-256 is not a valid encryptionalgorithm", e);
```
jverbraeken: Incorrect error description

app/src/main/java/com/nervousfish/nervousfish/modules/cryptography/EncryptedSaver.java
```diff
@@ -0,0 +1,266 @@
...
+        } catch (UnsupportedEncodingException e) {
+            LOGGER.error("UTF-8 is not a valid encoding method", e);
```
jverbraeken: Incorrect error description

app/src/main/java/com/nervousfish/nervousfish/modules/cryptography/EncryptedSaver.java
```diff
@@ -0,0 +1,266 @@
...
+        }
+        return null;
```
jverbraeken: Never return null, return a custom exception (see the exceptions package) if this line is reached instead

jverbraeken: Otherwise we can get NullPointerExceptions

app/src/main/java/com/nervousfish/nervousfish/modules/cryptography/EncryptedSaver.java
```diff
@@ -0,0 +1,266 @@
...
+     * @return  The encrypted string.
+     */
+    public static String hashWithoutSalt(final String pass) {
+        try {
+            final MessageDigest digest = MessageDigest.getInstance("SHA-256");
+            digest.reset();
+            final byte[] hash = digest.digest(pass.getBytes(StandardCharsets.UTF_8));
+            LOGGER.info("Hashed the pass with SHA-256, no salt");
+            return new String(hash, UTF_8);
+        } catch (NoSuchAlgorithmException e) {
```
jverbraeken: Should be final like all other exceptions. If you click "Cleanup code" this is done automatically for you (in this file at least)

app/src/main/java/com/nervousfish/nervousfish/modules/cryptography/EncryptedSaver.java
```diff
@@ -0,0 +1,266 @@
...
+    }
+
+    /**
+     * Hashes a pass to an encrypted string with a salt bytestring.
+     * @param salt The salt bytestring to encrypt the pass with.
+     * @param pass  The string to encrypt
+     * @return The encrypted string.
+     */
+    /*
+    public static String hashUsingSalt(byte[] salt, String pass){
```
jverbraeken: Again commented code

app/src/main/java/com/nervousfish/nervousfish/modules/cryptography/EncryptedSaver.java
```diff
@@ -0,0 +1,266 @@
...
+import javax.crypto.SecretKey;
+import javax.crypto.SecretKeyFactory;
+import javax.crypto.spec.PBEKeySpec;
+import javax.crypto.spec.PBEParameterSpec;
+
+/**
+ * Class used for encrypting strings.
+ * Commented out a bunch of methods in case of future needs.
+ */
+public final class EncryptedSaver {
```
jverbraeken: Can you move all methods from this class to EncryptorAdapter? That class is made to contain all these methods :)

app/src/main/java/com/nervousfish/nervousfish/modules/cryptography/EncryptedSaver.java
```diff
@@ -0,0 +1,266 @@
...
+        try {
+            final PBEKeySpec keySpec = new PBEKeySpec(password.toCharArray());
+            final SecretKeyFactory keyFactory = SecretKeyFactory.getInstance(PBE_WITH_MD5_AND_DES);
+            final SecretKey key = keyFactory.generateSecret(keySpec);
+            LOGGER.info("Key successfully made");
+            return key;
+        } catch (NoSuchAlgorithmException e) {
+            LOGGER.error("There isn't an algorithm like PBEWithMD5AndDES", e);
+        }
+        return null;
```
jverbraeken: Also here, can you throw a custom exception instead?

app/src/main/java/com/nervousfish/nervousfish/modules/cryptography/EncryptedSaver.java
jverbraeken: Also, the methods shouldn't be static imo. That creates very tight coupling to a certain class.

app/src/main/java/com/nervousfish/nervousfish/modules/cryptography/EncryptedSaver.java
```diff
@@ -0,0 +1,266 @@
...
+    }
+
+    /**
+     * Encrypts or decrypts a string with a password to
+     * @param toEncrypt The string to be encrypted/decrypted
+     * @param key  The secret key made from the password.
+     * @param encrypt   whether we're encrypting or decrypting.
+     * @return  The encrypted/decrypted bytearray.
+     */
+    public static String encryptOrDecryptWithPassword(final String toEncrypt, final SecretKey key, final boolean encrypt)
```
jverbraeken: Can you split this up into 2 methods, namely encrypt and decrypt? I think that "Password can be better omitted" because I want to use these methods for encrypting/decrypting the Bluetooth traffic as well, and those methods do not use a password but for example a rhythm.

jverbraeken: Would be even better if you removed the method makeKey, because the client now has to call 2 methods. First makeKey, the encryptOrDecryptWithPassword. As a client I would like to use your library by calling encrypt(byte[]) and decrypt(byte[]), because then I only have to use a single function, the code needed to generate a SecretKey becomes simpler (see https://stackoverflow.com/questions/14204437/convert-byte-array-to-secret-key) and because the password turns out to be a String accidentally, but the rhythm and visual verification are both other data types - and all data types can be turned into byte array easily (eg "asdf'.getBytes())

app/src/main/java/com/nervousfish/nervousfish/modules/cryptography/EncryptedSaver.java
```diff
@@ -0,0 +1,266 @@
...
+            LOGGER.error("There isn't an algorithm as PBEWithMD5AndDES", e);
+
+        } catch (NoSuchPaddingException e) {
+            LOGGER.error("There isn't a padding as PBEWithMD5AndDES", e);
+        } catch (InvalidKeyException e) {
+            LOGGER.error("The key is invalid", e);
+        } catch (InvalidAlgorithmParameterException e) {
+            LOGGER.error("The algorithm parameter is invalid", e);
+        }
+        return null;
```
jverbraeken: null

app/src/main/java/com/nervousfish/nervousfish/modules/cryptography/EncryptedSaver.java
```diff
@@ -0,0 +1,266 @@
...
+
+    /**
+     * Encrypt data using a public key.
+     * @param text  -   The data to encrypt
+     * @param publicKey -   The public key to encrypt the data with
+     * @return  The encrypted data in String.
+     * @throws IOException  -   Throws an IOException when the key can't be split up correctly
+     */
+    /*
+    public static String encryptUsingRSA(final String text, final IKey publicKey) throws IOException, InvalidKeySpecException,
```
jverbraeken: commented code

app/src/main/java/com/nervousfish/nervousfish/modules/database/GsonDatabaseAdapter.java
```diff
@@ -184,161 +181,217 @@ public boolean contactExists(final String name) throws IOException {
...
-    public void addProfile(final Profile profile) throws IOException {
-        // Get the list of profiles and add the new profile
-        final List<Profile> profiles = this.getProfiles();
-        profiles.add(profile);
-
-        final GsonBuilder gsonBuilder = new GsonBuilder()
-                .registerTypeHierarchyAdapter(IKey.class, new GsonKeyAdapter());
-        final Gson gsonParser = gsonBuilder.create();
+    public void updateProfile(final Profile newProfile) throws IOException {
+        getDatabase().setProfile(newProfile);
```
jverbraeken: You don't have to use getters internally, because a class is allowed to access his own variables

app/src/main/java/com/nervousfish/nervousfish/modules/database/GsonDatabaseAdapter.java
```diff
@@ -184,161 +181,217 @@ public boolean contactExists(final String name) throws IOException {
...
+        updateDatabase();
```
jverbraeken: this. (Hover -> Cleanup Code)

app/src/main/java/com/nervousfish/nervousfish/modules/database/GsonDatabaseAdapter.java
```diff
@@ -184,161 +181,217 @@ public boolean contactExists(final String name) throws IOException {
...
+        try {
+            final GsonBuilder gsonBuilder = new GsonBuilder().registerTypeHierarchyAdapter(IKey.class, new GsonKeyAdapter());
+            final Gson gsonParser = gsonBuilder.create();
 
-        // Throw if the contact to remove is not found
-        if (profiles.size() == lengthBefore) {
-            throw new IllegalArgumentException(CONTACT_NOT_FOUND);
+            final String databaseJson = gsonParser.toJson(getDatabase());
+            final String encryptedDatabase = EncryptedSaver.encryptOrDecryptWithPassword(databaseJson, getEncryptionKey(), true);
+            final Writer writer = this.fileSystem.getWriter(databasePath);
```
jverbraeken:
```
try (Writer writer = this.fileSystem.gerWriter(databasePath)) {
writer.write(encryptedDatabase);
}
```
If you do that, you don't have to use close(). It's safer to get accustomed to always using try() instead of close, because with try you cannot forget to write close(), resulting in a memory leak

app/src/main/java/com/nervousfish/nervousfish/modules/database/GsonDatabaseAdapter.java
```diff
@@ -184,161 +181,217 @@ public boolean contactExists(final String name) throws IOException {
...
         final GsonBuilder gsonBuilder = new GsonBuilder().registerTypeHierarchyAdapter(IKey.class, new GsonKeyAdapter());
         final Gson gsonParser = gsonBuilder.create();
+        helper.checkPassword(password);
 
-        // Update the database
-        final Writer writer = this.fileSystem.getWriter(this.profilesPath);
-        gsonParser.toJson(profiles, writer);
-        writer.close();
+        try {
+
```
jverbraeken: empty line

app/src/main/java/com/nervousfish/nervousfish/modules/database/GsonDatabaseAdapter.java
```diff
@@ -184,161 +181,217 @@ public boolean contactExists(final String name) throws IOException {
...
+            final String databaseFileString = helper.readDatabaseToString();
```
jverbraeken: Nice coding style!

app/src/main/java/com/nervousfish/nervousfish/modules/database/GsonDatabaseAdapter.java
```diff
@@ -184,161 +181,217 @@ public boolean contactExists(final String name) throws IOException {
...
-        final int lengthBefore = profiles.size();
-        profiles.remove(oldProfile);
-
-        // Throw if the contact to update is not found
-        if (profiles.size() == lengthBefore) {
-            throw new IllegalArgumentException(CONTACT_NOT_FOUND);
-        }
+    public void createDatabase(final Profile profile, final String password) throws IOException {
+        LOGGER.info("Started creating database for profile");
+        initializePassword(password);
```
jverbraeken: this (should already have been cleanup by now)

app/src/main/java/com/nervousfish/nervousfish/modules/database/GsonDatabaseAdapter.java
```diff
@@ -184,161 +181,217 @@ public boolean contactExists(final String name) throws IOException {
...
-        final Writer writer = this.fileSystem.getWriter(this.profilesPath);
-        gsonParser.toJson(profiles, writer);
-        writer.close();
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public boolean checkFirstUse() {
+        final File filePass = new File(passwordPath);
+        final File fileDb = new File(databasePath);
```
jverbraeken: Can you add a new method to AndroidFileSystemAdapter called checkFileExists(String path) and call that function instead?

jverbraeken: The database adapter shouldn't be responsible for file related stuff

app/src/main/java/com/nervousfish/nervousfish/modules/database/GsonDatabaseAdapter.java
```diff
@@ -184,161 +181,217 @@ public boolean contactExists(final String name) throws IOException {
...
+        return filePass.exists() && fileDb.exists();
     }
 
     /**
      * {@inheritDoc}
      */
     @Override
-    public String getUserPassword() throws IOException {
-        return null;
+    public void deleteDatabase() {
```
jverbraeken: A method like deleteFile(final String path) in AndroidFileSystemAdapter would better suit the intended functionality of this class

app/src/main/java/com/nervousfish/nervousfish/modules/database/GsonDatabaseAdapter.java
```diff
@@ -184,161 +181,217 @@ public boolean contactExists(final String name) throws IOException {
...
+        databaseMap.put(DATABASE_PASS, new DatabasePass(password));
+
+
+        final DatabasePass databasePass = getDatabasePass();
+        final GsonBuilder gsonBuilder = new GsonBuilder().registerTypeHierarchyAdapter(IKey.class, new GsonKeyAdapter());
+        final Gson gsonParser = gsonBuilder.create();
+        final String databasePassJson = gsonParser.toJson(databasePass);
+        LOGGER.info("Database pass: " + databasePassJson);
+
+        final Writer writer = this.fileSystem.getWriter(passwordPath);
```
jverbraeken: try(

app/src/main/java/com/nervousfish/nervousfish/modules/database/GsonDatabaseAdapter.java
```diff
@@ -184,161 +181,217 @@ public boolean contactExists(final String name) throws IOException {
...
-    private void ensureClassInvariant() throws InvalidObjectException {
-        assertNotNull(this.contactsPath);
-        assertNotNull(this.profilesPath);
-        assertNotNull(this.fileSystem);
+    private SecretKey getEncryptionKey() throws IOException {
+        final Object object = databaseMap.get(ENCRYPTION_KEY);
+        if (object instanceof SecretKey) {
+            return (SecretKey) object;
+        } else {
+            throw new IOException(DATABASE_NOT_CREATED);
```
jverbraeken: Can you create a new class extending RuntimeException in the package exceptions? IOException is too vague (actually, it has little to do with IO at all)

app/src/main/java/com/nervousfish/nervousfish/modules/database/GsonDatabaseAdapterHelper.java
```diff
@@ -0,0 +1,86 @@
...
+
+    /**
+     * Reads the database to a string.
+     * @return A String containing the json database
+     */
+    String readDatabaseToString() throws IOException {
+        // Get the database from the database file
+        final BufferedReader databaseReader = (BufferedReader) this.fileSystem.getReader(databasePath);
+
+        final StringBuffer databaseFileStringBuffer = new StringBuffer();
```
jverbraeken: I don't think that multiple threads will access this method at the same time, so StringBuilder (which is faster although unsynchronized) is a better choice

app/src/main/java/com/nervousfish/nervousfish/modules/database/GsonDatabaseAdapterHelper.java
```diff
@@ -0,0 +1,86 @@
...
+
+    /**
+     * Checks the password, throws IOException if it's wrong.
+     * @param password  -   The password to check.
+     * @throws IOException Thrown when password is wrong.
+     */
+    void checkPassword(final String password) throws IOException {
+        final GsonBuilder gsonBuilder = new GsonBuilder().registerTypeHierarchyAdapter(IKey.class, new GsonKeyAdapter());
+        final Gson gsonParser = gsonBuilder.create();
+        final BufferedReader passReader = (BufferedReader) this.fileSystem.getReader(passwordPath);
```
jverbraeken: try(....) {
}

app/src/main/java/com/nervousfish/nervousfish/modules/database/GsonDatabaseAdapterHelper.java
```diff
@@ -0,0 +1,86 @@
...
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.Serializable;
+import java.lang.reflect.Type;
+
+/**
+ * Helper method for the logical functionality of the {@link GsonDatabaseAdapter}.
+ */
+class GsonDatabaseAdapterHelper implements Serializable {
```
jverbraeken: I think that the name is too vague. Can you rename this class to GsonDatabaseAdapterLoader, move the contents of the method loadDatabase from GsonDatabaseAdapter to this class and let the old method call this new one?

jverbraeken: All other methods can be private after that is done

app/build.gradle
```diff
@@ -37,6 +37,9 @@ android {
         }
     }
 
+
+
+
```
clenrock: Could you delete these extra enters?

app/src/androidTest/java/com/nervousfish/nervousfish/test/CreateProfileSteps.java
```diff
@@ -60,11 +89,11 @@ public void iAmViewingTheCreateProfileActivity() throws IOException {
...
-        final List<Profile> profiles = serviceLocator.getDatabase().getProfiles();
-        for (final Profile profile : profiles) {
-            serviceLocator.getDatabase().deleteProfile(profile);
-        }
-        assertTrue(serviceLocator.getDatabase().getProfiles().isEmpty());
+        //final List<Profile> profiles = serviceLocator.getDatabase().getProfiles();
+        //for (final Profile profile : profiles) {
+        //    serviceLocator.getDatabase().deleteProfile(profile);
+        //}
+       // assertTrue(serviceLocator.getDatabase().getProfiles().isEmpty());
```
clenrock: Could you remove the commented code, or alter the test

app/src/androidTest/java/com/nervousfish/nervousfish/test/CreateProfileSteps.java
```diff
@@ -144,8 +173,8 @@ public void repeatPasswordInputFieldShouldBeRed() {
 
     @Then("^the profile with (.*?) should be saved in the database$")
     public void profileWithNameShouldBeInDatabase(final String name) throws IOException {
-        final List<Profile> profiles = serviceLocator.getDatabase().getProfiles();
-        assertEquals(name, profiles.get(0).getName());
+        //final List<Profile> profiles = serviceLocator.getDatabase().getProfiles();
+//        assertEquals(name, profiles.get(0).getName());
```
clenrock: Same

app/src/main/java/com/nervousfish/nervousfish/activities/LoginActivity.java
```diff
@@ -34,15 +34,9 @@ protected void onCreate(final Bundle savedInstanceState) {
...
         this.setContentView(R.layout.login);
 
-        final IServiceLocator serviceLocator = NervousFish.getServiceLocator();
-        final IDatabase database = serviceLocator.getDatabase();
-        try {
-            this.actualPassword = database.getUserPassword();
-        } catch (final IOException e) {
-            LOGGER.error("Failed to retrieve password from database", e);
-        }
+        serviceLocator = NervousFish.getServiceLocator();
```
clenrock: Can you add: this.

app/src/main/java/com/nervousfish/nervousfish/modules/constants/Constants.java
```diff
@@ -24,6 +24,8 @@
     private static final Logger LOGGER = LoggerFactory.getLogger("Constants");
     private static final String DB_USERDATA_PATH = "accountInformation.json";
     private static final String DB_CONTACTS_PATH = "contacts.json";
+    private static final String DB_DATABASE_PATH = "database.txt";
```
clenrock: Maybe "database" is a bit of a confusing name, we will talk about this in the meeting

app/src/main/java/com/nervousfish/nervousfish/modules/cryptography/EncryptedSaver.java
```diff
@@ -0,0 +1,266 @@
...
+    private static final int SEED = 1234569;
+    private static final int IV_SPEC_SIZE = 8;
+    private static final String UTF_8 = "UTF-8";
+
+    /**
+     * Unused constructor for utility class.
+     */
+    private EncryptedSaver() {
+        // Prevent instantiations of QRGenerator
+
```
clenrock: Unused enter

app/src/main/java/com/nervousfish/nervousfish/modules/cryptography/EncryptedSaver.java
```diff
@@ -0,0 +1,266 @@
...
+            }
+        } catch (UnsupportedEncodingException e) {
+            LOGGER.error("UTF-8 is no longer an encoding algorithm", e);
+            return null;
+        }
+    }
+
+
+
+
```
clenrock: A lot of enters

app/src/main/java/com/nervousfish/nervousfish/modules/database/GsonDatabaseAdapterHelper.java
```diff
@@ -0,0 +1,86 @@
...
+
+    /**
+     * Helper class for GsonDatabaseAdapter.
+     * @param serviceLocator Can be used to get access to other modules
+     */
+    GsonDatabaseAdapterHelper(final IServiceLocator serviceLocator) {
+        this.fileSystem = serviceLocator.getFileSystem();
+        this.databasePath = serviceLocator.getConstants().getDatabasePath();
+        this.passwordPath = serviceLocator.getConstants().getPasswordPath();
+
```
clenrock: Unused enter

app/src/main/java/com/nervousfish/nervousfish/modules/database/GsonDatabaseAdapterHelper.java
```diff
@@ -0,0 +1,86 @@
...
+                break;
+            }
+            passwordFileStringBuffer.append(line);
+        }
+        final String passwordFileString = passwordFileStringBuffer.toString();
+        final DatabasePass databasePass = gsonParser.fromJson(passwordFileString, TYPE_DATABASE_PASS);
+        if (!databasePass.getEncryptedPassword().equals(EncryptedSaver.hashWithoutSalt(password))) {
+            throw new IOException("Password is wrong " + databasePass.getEncryptedPassword() + " " + EncryptedSaver.hashWithoutSalt(password));
+        }
+        passReader.close();
```
clenrock: Is this reader not closed when an exception is thrown? Shouldn't this always be closed?

app/src/test/java/com/nervousfish/nervousfish/modules/database/GsonDatabaseAdapterTest.java
```diff
@@ -82,14 +94,38 @@ public void testConstructorNotFailsWhenFileInitializationFails() {
         assertNotNull(database);
     }
 
+/*    @Test
+    public void testAddContactWithSingleKey() throws Exception {
+
+
+        final GsonBuilder gsonBuilder = new GsonBuilder().registerTypeHierarchyAdapter(IKey.class, BaseTest.accessConstructor(GsonKeyAdapter.class));
+        final Gson gsonParser = gsonBuilder.create();
+
```
clenrock: Fix this test or delete it

app/src/main/java/com/nervousfish/nervousfish/modules/database/GsonDatabaseAdapterHelper.java
jverbraeken: If we use try() {} we don't even have to worry about closing the stream

app/src/main/java/com/nervousfish/nervousfish/data_objects/Database.java
KilianCallebaut: Profile is Serializable though?

app/src/main/java/com/nervousfish/nervousfish/modules/cryptography/EncryptedSaver.java
KilianCallebaut: Yeah I mentioned this in the pr description, it's so that the working methods would still be there in case anyone needs them for later use

app/src/main/java/com/nervousfish/nervousfish/modules/cryptography/EncryptedSaver.java
KilianCallebaut: How so, it's what it would imply when that error is thrown.

app/src/main/java/com/nervousfish/nervousfish/modules/cryptography/EncryptedSaver.java
KilianCallebaut: again it would imply that if that error is thrown

app/src/main/java/com/nervousfish/nervousfish/modules/cryptography/EncryptedSaver.java
KilianCallebaut: see previous

app/src/main/java/com/nervousfish/nervousfish/modules/cryptography/EncryptedSaver.java
KilianCallebaut: That's kind of why I kept in the rsa key algorithms, seeing that they are much safer for use in datatransport. They are completely working but commented out because I didn't need them and they were causing checkstyle errors. If you really want to use this though I will do what you say.

app/src/main/java/com/nervousfish/nervousfish/modules/database/GsonDatabaseAdapter.java
KilianCallebaut: This getter is used to ensure that there is a database present in the databasemap, or in other words that the database has been created and loaded

app/src/test/java/com/nervousfish/nervousfish/modules/database/GsonDatabaseAdapterTest.java
ericcornelissen: Fix this test ~~or delete it~~

ftfy

app/src/main/java/com/nervousfish/nervousfish/modules/cryptography/EncryptedSaver.java
KilianCallebaut: Also the generatekey thing is kind of what keeps it safe and allows for writes without having to constantly input a password. Without this I would either have to save the unencrypted password or encrypt using the encrypted password, both making the program less safe and my work kind of obsolete

app/src/main/java/com/nervousfish/nervousfish/modules/database/GsonDatabaseAdapterLoader.java
```diff
@@ -0,0 +1,134 @@
...
+        passReader.close();
+    }
+
+    /**
+     * Loads the Secretkey from a given password.
+     *
+     * @param password THe password to load a key from
+     * @return The secretKey made from the password\
+     */
+    public SecretKey loadKey(final String password) throws IOException, InvalidKeySpecException {
```
ericcornelissen: Why is this part of the database, shouldn't this be done separately from the database before asking the database for its content?

app/src/main/java/com/nervousfish/nervousfish/service_locator/EntryActivity.java
```diff
@@ -42,19 +35,12 @@ protected void onResume() {
...
+        final Intent intent;
 
-        final Intent intent = new Intent();
-        if (profiles.isEmpty()) {
-            intent.setComponent(new ComponentName(this, WelcomeActivity.class));
+        if (serviceLocator.getDatabase().checkFirstUse()) {
+            intent = new Intent(this, LoginActivity.class);
         } else {
-            intent.setComponent(new ComponentName(this, LoginActivity.class));
+            intent = new Intent(this, WelcomeActivity.class);
```
ericcornelissen: Can you restore the code as it was before (with the `final Intent intent = new Intent();`)

app/src/main/java/com/nervousfish/nervousfish/modules/constants/Constants.java
KilianCallebaut: never talked about this, can we accept it as is?

app/src/main/java/com/nervousfish/nervousfish/data_objects/Database.java
```diff
@@ -0,0 +1,106 @@
...
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * A database object to write and encrypt to file
+ */
+public final class Database implements Serializable {
+    private static final long serialVersionUID = -4715364587956219157L;
```
TheBonheurs: Should these be the same? See below

app/src/main/java/com/nervousfish/nervousfish/data_objects/KeyPair.java
```diff
@@ -1,10 +1,17 @@
...
+import java.io.ObjectInputStream;
+import java.io.Serializable;
+
 /**
  * KeyPair POJO which keeps a public and private key.
  */
-public final class KeyPair {
+public final class KeyPair implements Serializable {
 
+    private static final long serialVersionUID = -4715364587956219157L;
```
TheBonheurs: Here, the same serialization uuid !

app/src/main/java/com/nervousfish/nervousfish/data_objects/Profile.java
```diff
@@ -1,54 +1,79 @@
...
 /**
- * A Profile POJO to store a name, public key and private key.
+ * A Profile POJO to store a contact representing the user and the user's keypairs.
  */
-public class Profile {
+public class Profile implements Serializable {
 
-    private final String name;
-    private final List<KeyPair> keyPairs = new ArrayList<>();
+    private static final long serialVersionUID = -4715364587956219157L;
```
TheBonheurs: Again

app/src/main/java/com/nervousfish/nervousfish/modules/cryptography/EncryptorAdapter.java
```diff
@@ -36,4 +64,132 @@ private EncryptorAdapter(final IServiceLocator serviceLocator) {
...
+
+    /**
+     * Method that return a configured {@link Cipher}.<br>
+     * It set the ivSpec and the encryption to
+     * PBEWithMD5AndDES
+     *
+     * @param key    The SecretKey used to encrypt
+     * @param ivSpec The initializations vector for the encryption
+     * @param mode   Whether we're encrypting or decrypting
+     * @return the configured Cipher.
```
TheBonheurs: {@link cipher}.
It sets...
Method that returns ...

app/src/main/java/com/nervousfish/nervousfish/modules/cryptography/IEncryptor.java
```diff
@@ -1,10 +1,55 @@
...
 /**
  * Defines a module used for encrypting / decrypting messages that can be used by a service locator.
  */
 
 public interface IEncryptor extends IModule {
+
+    /**
+     * Hashes a given pass to an encrypted string that can't be decrypted.
+     *
+     * @param pass THe string to encrypt
```
TheBonheurs: The...

app/src/main/java/com/nervousfish/nervousfish/modules/database/IDatabase.java
```diff
@@ -25,7 +25,8 @@
      *
      * @param contactName The {@link String} with the name of the contact to remove from the database.
      */
-    void deleteContact(String contactName) throws IllegalArgumentException, IOException;
+    void deleteContact(String contactName) throws IOException;
+
```
TheBonheurs: Why this whitespace?

app/src/main/java/com/nervousfish/nervousfish/data_objects/Profile.java
```diff
@@ -1,54 +1,79 @@
...
      *
-     * @param name the name belonging to the Profile.
-     * @param keyPair the public/private key-pair.
+     * @param contact  The contact belonging to the user.
+     * @param keyPairs the public/private key-pairs of the user.
      */
-    public Profile(final String name, final KeyPair keyPair) {
-        this.name = name;
-        this.keyPairs.add(keyPair);
+    public Profile(final Contact contact, final List<KeyPair> keyPairs) {
```
KilianCallebaut: See Joost's reworking of the database

KilianCallebaut: The contact represents the user himself

ericcornelissen: Where should I see that? (Still don't agree though, at best you can add a method to the Profile class that transforms it into a Contact :wink:)

app/src/main/java/com/nervousfish/nervousfish/activities/CreateProfileHelper.java
```diff
@@ -26,21 +26,48 @@
...
+
+    /**
      * Generates a KeyPair based on the type selected.
      *
      * @param keyType The type of key to generate.
      * @return a {@link KeyPair} with the key type selected
      */
-    Collection<KeyPair> generateKeyPairs(final IKey.Types keyType) {
-        final Collection<KeyPair> keyPairs = new ArrayList<>();
+    List<KeyPair> generateKeyPairs(final IKey.Types keyType) {
```
KilianCallebaut: The fact that I already build it around a list :) Also lists are for sure serializable and won't give any trouble writing to file, but I don't know for sure if that isn't the case with collections.

app/src/main/java/com/nervousfish/nervousfish/data_objects/KeyPair.java
```diff
@@ -72,4 +79,48 @@ public int hashCode() {
         return this.name.hashCode() + this.publicKey.hashCode() + this.privateKey.hashCode();
     }
 
+    /**
+     * Serialize the created proxy instead of this instance.
+     */
+    private Object writeReplace() {
+        return new KeyPair.SerializationProxy(this);
```
KilianCallebaut: I don't know actually I added them by example of other serializable objects that were already made

ericcornelissen: In that case you can remove them 🙂

app/src/main/java/com/nervousfish/nervousfish/activities/CreateProfileHelper.java
ericcornelissen: > The fact that I already build it around a list

Well, you might have to rethink your code then


> Also lists are for sure serializable and won't give any trouble writing to file, but I don't know for sure if that isn't the case with collections.

Luckily for you `Collection`s are not serializable, so I'll give you a pass on this one 😉

app/src/main/java/com/nervousfish/nervousfish/data_objects/Database.java
```diff
@@ -0,0 +1,106 @@
...
+     */
+    public Profile getProfile() {
+        return this.profile;
+    }
+
+    /**
+     * Sets the user's profile to a new profile
+     * @param profile The new profile.
+     */
+    public void setProfile(final Profile profile) { this.profile = profile; }
```
KilianCallebaut: Thing is, for a method like update profile it will require this. I have 2 alternatives: 1 make a sort of wrapper around the profile so that I can reset it or 2 make the variables of profile alterable so that it's contact object and it's list of keypairs can be updated.

app/src/main/java/com/nervousfish/nervousfish/modules/constants/Constants.java
```diff
@@ -54,6 +56,23 @@ private Constants(final IServiceLocator serviceLocator) {
         return new ModuleWrapper<>(new Constants(serviceLocator));
     }
 
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public String getPasswordPath() {
```
KilianCallebaut: In principle there shouldn't be an unencrypted database anymore, but I guess we'll discuss that on the next meeting

app/src/main/java/com/nervousfish/nervousfish/modules/cryptography/EncryptorAdapter.java
```diff
@@ -1,17 +1,45 @@
...
+
 /**
  * An adapter to the default Java class for encrypting messages
  */
 public final class EncryptorAdapter implements IEncryptor {
     private static final long serialVersionUID = 5930930748980177440L;
     private static final Logger LOGGER = LoggerFactory.getLogger("EncryptorAdapter");
+    private static final String PBE_WITH_MD5_AND_DES = "PBEWithMD5AndDES";
+    private static final String UTF_8_NO_LONGER_SUPPORTED = "UTF-8 is no longer an encoding algorithm";
+    private static final String CANNOT_HAPPEN_UTF_8 = "Cannot happen, no encodign algorithm like UTF-8";
```
KilianCallebaut: Avoiding checkstyle errors. These are used in exactly the same way so Checkstyle told me to make a variable out of them.

app/src/main/java/com/nervousfish/nervousfish/modules/cryptography/EncryptorAdapter.java
```diff
@@ -36,4 +64,132 @@ private EncryptorAdapter(final IServiceLocator serviceLocator) {
         return new ModuleWrapper<>(new EncryptorAdapter(serviceLocator));
     }
 
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public String hashWithoutSalt(final String pass) throws EncryptionException {
```
KilianCallebaut: any string. The use case is that you can check whether or not a password is correct, without anyone outside of the user being able to read the password. If you save an encrypted string like this using your password and encrypt that same password like this later, it will hash to the same string, meaning you entered the same password.

KilianCallebaut: The fact that it says password is because in practice this is about the only thing you can really use it for.

app/src/main/java/com/nervousfish/nervousfish/modules/database/GsonDatabaseAdapterLoader.java
KilianCallebaut: My reasoning was so that it would be clear that you were loading the database key but sure where would you suggest that I'd move it to?

app/src/main/java/com/nervousfish/nervousfish/modules/database/GsonDatabaseAdapterLoader.java
```diff
@@ -0,0 +1,134 @@
...
+        return databaseFileStringBuilder.toString();
+    }
+
+    /**
+     * Checks the password, throws IOException if it's wrong.
+     *
+     * @param password -   The password to check.
+     * @throws IOException Thrown when password is wrong.
+     */
+    private void checkPassword(final String password) throws IOException {
```
KilianCallebaut: Thing is in its current state it would never be able to return false since when it's false it throws an IOException, making it easier for debugging because it will return the difference in encrypted strings

app/src/main/java/com/nervousfish/nervousfish/data_objects/Profile.java
KilianCallebaut: https://drive.google.com/file/d/0Bx7YFLoLhW0Iei1tVmY1ekZYOTQ/view

ericcornelissen: Ok, but:
> Still don't agree though, at best you can add a method to the Profile class that transforms it into a Contact 😉

app/src/main/java/com/nervousfish/nervousfish/data_objects/Database.java
ericcornelissen: > This is kind of a dangerous practice, isn't it possible to set the profile in the constructor

Why is this not possible?

app/src/main/java/com/nervousfish/nervousfish/modules/constants/Constants.java
ericcornelissen: Regardless...
> why don't you store this in a file with the same name?

app/src/main/java/com/nervousfish/nervousfish/modules/cryptography/EncryptorAdapter.java
ericcornelissen: I might be mistaken, but these all seem like messages you use for logging? I would suggest making them less generic and putting them back into their respective log statement

app/src/main/java/com/nervousfish/nervousfish/modules/database/GsonDatabaseAdapterLoader.java
ericcornelissen: I guess you should be able to obtain the secret key from an `IEncryptor` instance whenever you need it

app/src/main/java/com/nervousfish/nervousfish/modules/cryptography/EncryptorAdapter.java
ericcornelissen: Good use case, forgot about that 😅 In that case, can you rename the method either to `hashPassword` or to "`hashString` and change the parameter name to `string`"

app/src/main/java/com/nervousfish/nervousfish/modules/database/GsonDatabaseAdapterLoader.java
ericcornelissen: I agree with @TheBonheurs 


> making it easier for debugging because it will return the difference in encrypted strings

That is not true, the difference in these string should not tell you anything about the original strings (given they're properly hashed)

app/src/main/java/com/nervousfish/nervousfish/modules/constants/Constants.java
kcallebaut: Same name as what?

app/src/main/java/com/nervousfish/nervousfish/data_objects/Profile.java
kcallebaut: So that there's a constructor with only the name?

app/src/main/java/com/nervousfish/nervousfish/modules/database/GsonDatabaseAdapterLoader.java
kcallebaut: So i should add a boolean return doing effectively nothing because it would throw an exception beforehand?

app/src/main/java/com/nervousfish/nervousfish/modules/cryptography/EncryptorAdapter.java
kcallebaut: But they're exactly the same logging problems that shouldn't happen in the first place unless utf-8 and pbe with md5 are no longer algorithms. The one before is for making the key which is correct

app/src/main/java/com/nervousfish/nervousfish/data_objects/Database.java
kcallebaut: I can make a new databaseobject now that i think about it when updating the profile

app/src/main/java/com/nervousfish/nervousfish/modules/database/GsonDatabaseAdapterLoader.java
kcallebaut: Gson databaseloader is a helper of the implementing iencryptor class so that's covered

app/src/main/java/com/nervousfish/nervousfish/modules/database/GsonDatabaseAdapterLoader.java
TheBonheurs: I think you can still, albeit not much, get something out of a true return value right?

KilianCallebaut: Ok then I'll just throw the exception in the loadkey method

app/src/main/java/com/nervousfish/nervousfish/data_objects/Profile.java
KilianCallebaut: done

ericcornelissen: I was thinking more along the lines of:
```java
public class Profile implements Serializable {

    private static final long serialVersionUID = 8191245914949893284L;
    private final String name;
    private final List<KeyPair> keyPairs;

    /**
     * The constructor for the {@link Profile} class.
     *
     * @param name     The name for the profile/of the user.
     * @param keyPairs The public/private key-pairs of the user.
     */
    public Profile(final String name final List<KeyPair> keyPairs) {
        this.name = name;
        this.keyPairs = keyPairs;
    }

    /**
     * The constructor for the {@link Profile} class.
     *
     * @param name  The contact belonging to the user.
     * @param keyPairs the public/private key-pairs of the user.
     */
    public Profile(final String name, final List<KeyPair> keyPairs) {
        this.name = name;
        this.keyPairs = keyPairs;
    }

    /**
     * Adds a new keyPair to the profile
     *
     * @param keyPair the keyPair to add.
     */
    public void addKeyPair(final KeyPair keyPair) {
        keyPairs.add(keyPair);
    }

    /**
     * Returns the profile as a {@link Contact}.
     *
     * @return - The Contact POJO.
     */
    public Contact getContact() {
        return new Contact(name, publicKeys);
    }

    /**
     * Returns the Keypairs of the user.
     *
     * @return The list of keypairs of the user.
     */
    public List<KeyPair> getKeyPairs() {
        return this.keyPairs;
    }

    /**
     * Returns the name of the profile.
     *
     * @return Name of the profile.
     */
    public String getName() {
        return this.name;
    }

}
```

app/src/main/java/com/nervousfish/nervousfish/modules/constants/Constants.java
ericcornelissen: Why don't the encrypted and non-encrypted version of the database don't have the same filename?

app/src/main/java/com/nervousfish/nervousfish/modules/cryptography/EncryptorAdapter.java
ericcornelissen: A few points:
- There are typos in these strings.
- You're not properly using them, for example on [line 84](https://github.com/ericcornelissen/NervousFish/pull/252/files/65ae0f3a84342b70782e9ba8a6ae88a5f37d105d#diff-4a0ae8e2bf0ce74e2c03082d5ace1a3dR84) you have hardcoded the `CANNOT_HAPPEN_UTF_8` text.
- You're not consistently using them. For example, you don't have a string for the SHA-256 error.
- If you have to rewrite it so often you might consider refactoring so you have less duplicate code. E.g. for the UTF error:
```java
private String getKeyAsString(final byte[] key) {
     try {
         return new String(secretString, UTF_8);
     } catch (final UnsupportedEncodingException e) {
         LOGGER.error("the message", e);
         throw new EncryptionException("the message", e);
     }
}
```

app/src/main/java/com/nervousfish/nervousfish/modules/constants/Constants.java
ericcornelissen: @clenrock, @KilianCallebaut, what are we going to do about this?

app/src/main/java/com/nervousfish/nervousfish/modules/constants/Constants.java
ericcornelissen: My bad 😅 misunderstood what `getPasswordPath()` was supposed to do

app/src/main/java/com/nervousfish/nervousfish/modules/cryptography/EncryptorAdapter.java
KilianCallebaut: The sha-256 error can only be thrown at 1 place. Also it's in encryption and decryption so that method won't work

app/src/main/java/com/nervousfish/nervousfish/modules/constants/Constants.java
clenrock: I accepted it :)

ericcornelissen:
Merged regardless of Travis CI status, the build time exceeded time limit (50 min) multiple times

PR #259, opened 2017-06-15
Throw and test the Deserialization exception

jverbraeken:
- Relevant Issues: -
- Related Pull Requests: -

* * *

## What
This Pull Request
- lets DataReceiver throw a DeserializationException when something went wrong with the deserialization
- Tests the DeserializationException

## Why
This Pull Request is needed because no exception was thrown when something went wrong while deserializing

## How
This feature can be viewed/tested within the project by -

## Alternative implementation

### DIFF
--- a/app/src/main/java/com/nervousfish/nervousfish/modules/pairing/APairingHandler.java
+++ b/app/src/main/java/com/nervousfish/nervousfish/modules/pairing/APairingHandler.java
@@ -49,6 +49,7 @@ abstract class APairingHandler implements IPairingHandler {
                 this.serviceLocator.postOnEventBus(new NewDataReceivedEvent(object.getData(), object.getClazz()));
             } catch (final ClassNotFoundException | IOException e) {
                 LOGGER.error(" Couldn't start deserialization!", e);
+                throw new DeserializationException(e);
             }
         });
     }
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/nervousfish/nervousfish/modules/pairing/APairingHandler.java
```diff
@@ -49,6 +49,7 @@
                 this.serviceLocator.postOnEventBus(new NewDataReceivedEvent(object.getData(), object.getClazz()));
             } catch (final ClassNotFoundException | IOException e) {
                 LOGGER.error(" Couldn't start deserialization!", e);
+                throw new DeserializationException(e);
```
ericcornelissen: Are you 100% sure this is the only place where a `DeserializationException` should be thrown? It seems oddly specific for such a generic exception...

jverbraeken: The only place where objects are deserialized is after they are received from another device over network. All incoming network traffic is handled by DataReceiver from APairingHandler

ericcornelissen: OK 👍

PR #282, opened 2017-06-16
Remove the unused MultiContact

jverbraeken:
- Relevant Issues: -
- Related Pull Requests: -

* * *

## What
This Pull Request adds to the repository the removal of the unused file MultiContact

## Why
This Pull Request is needed because we don't want unused files in our repository

## How
This feature can be viewed/tested within the project by verifying that everything still works

## Alternative implementation

## Notes

Maybe the file is there because we may need it later. I don't think that it's a good idea to add things to the repository so much earlier before we need it, but okay, please say so if that's the reason the file's in the repo


ericcornelissen:
Merged regardless of Travis CI status, either the emulator wouldn't start or there were unexpected failing androidTests (likely due to a crashed emulator)

PR #283, opened 2017-06-16
Move DatabaseException to the database package

jverbraeken:
- Relevant Issues: -
- Related Pull Requests: -

* * *

## What
This Pull Request moves DatabaseException from exceptions to modules.database

## Why
This Pull Request is needed because we don't want to put interfaces in a package named interfaces, to put abstract class in a package named abstract classes and we also don't want to put all exceptions in a package called exceptions. We want to group things that are related to each other by functionality, because they depend on each other, in the same package. Only exceptions that are really generic can be justified to be in such a generic package like exceptions and DatabaseException certainly does not belong to the category.

## How
This feature can be viewed/tested within the project by 

## Alternative implementation


ericcornelissen:
Merged regardless of Travis CI status, Findbugs workers kept crashing 😕

Issue #273, opened 2017-06-16
Fix the timeout errors of Travis


jverbraeken:
- Parent issues: -
- Relevant issues: -
- Estimated effort: 5

* * *

## User story
As a developer I don't want to get timeout errors of Travis all the time so that I can verify that the project is still working on the repository

## What
This issue demands to fix the timeout errors of Travis (i.e. Waiting for emulator to start)

## Why
This issue is needed for the project because without Travis it's a lot harder to verify that branches are not broken

## How
This feature can be added by solving some way the timeout errors of Travis

## Definition of ready
This features is not ready until the following issues are finished:

- 

## Definition of done
This task can be considered done when

- Travis does (almost) never give timeout errors anymore when pushing to a branch

PR #290, opened 2017-06-18
Update Travis CI configuration 

ericcornelissen:
- Relevant Issues: #273 
- Related Pull Requests: -

* * *

## What

This Pull Request updates the Travis CI integration configuration significantly. It does so by having multiple builds running parallel, each performing a subtask of the build process, where builds that are know to fail frequently are allowed to fail (though it is always encouraged to get these builds working!)

For now there are four jobs, 2 of which are allowed to fail. First the 2 jobs that need to succeed
- **quality**: Runs `gradle checkstyle findbugs pmd lint` and is there to make sure the code quality is good.
- **unittest**: Runs `gradle testDebugUnitTest testReleaseUnitTest`, so the unittests (and not the androidTests!) for the project. This makes sure at least some of the code is working correctly.

The other 2 jobs are allowed to fail, these jobs depend on the emulator/androidTests which (as you certainly have experience) tend to fail for no apparent reason...
- **androidtest**: Runs `gradle connectedDebugAndroidTest`, so the androidTests for the project on a real emulator. Since it solely relies on the emulator, it is allowed to fail (for now).
- **coverals**: Runs `gradle jacocoTestReport coveralls`, so is intended to report the test coverage to Coveralls. As this is not *that* important (and relieson the emulator) it is allowed to fail.

In order to make this I use some scripts, which can be found in `~/gradle/scripts/`. One of them can be used to start an emulator for the job, one is executed as "before_script" in `.travis.yml`, and the last one is executed as "script" in `.travis.yml`.

Furthermore, I've tweaked the `.travis.yml` a little bit in hopes of improving performance & making the configuration file more readable 🙂 

## Why

This Pull Request is needed so we can start using Travis CI again! I know the solution is not perfect, but it is at least something! 😄 

## How
This feature can be viewed/tested within the project by clicking on the Travis CI status of this PR and see the details there. You can see that multiple jobs are running/have run and that some are allowed to fail.

## Alternative implementation

Other implementations that I've have considered are:
- Use a "build" job that just runs `gradle build`: Have chosen to have the "unittest" and "quality" jobs instead because it provides more insight into whether your build was successful (both can fail, but you would possibly only see one when using `gradle build`). And it does most of the things `gradle build` does.
- Don't allow the "androidtest" job to fail: Would love to have done that, but unfortunately the emulator on Travis is too unstable for this. I hope everyone runs `gradle connectedCheck` on their branch before opening a PR 😄 

## Notes

Again, I know the solution is not perfect, but at least we can rely on some continues integration now 🎉


gradle/scripts/travis_before_script.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+set -e
+
+# Starting emulators is very costly, only done for components that need it
+if [ "$COMPONENT" == "androidtest" ] || [ "$COMPONENT" == "coveralls" ]; then
```
TheBonheurs: nice one!

config/quality.gradle
```diff
@@ -49,7 +49,7 @@ task checkstyleTest(type: Checkstyle) {
     classpath = files()
 }
 
-task findbugs(type: FindBugs) {
+task findbugs(type: FindBugs, dependsOn: ['compileDebugSources', 'compileReleaseSources']) {
```
TheBonheurs: Does this ensure that the project is built/compliled every time before findbugs checks are applied?

ericcornelissen: Indeed, it compiles all the classes Findbugs want to analyze 🙂

gradle/scripts/travis_before_script.sh
ericcornelissen: 😇

.travis.yml
```diff
@@ -24,28 +35,34 @@ android:
...
+matrix:
+  fast_finish: true
+  allow_failures:
+    # Often the emulator doesn't start or crashes during execution
+    - env: COMPONENT=androidtest
+    # Coveralls is nice to have, but not a must have
+    - env: COMPONENT=coveralls
+
+git:
+  depth: 5
```
jverbraeken: Can you tell me why you chose 5?

gradle/scripts/travis_create_avd.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+# Creates and starts an emulator.
+android-update-sdk --components=sys-img-armeabi-v7a-android-22 --accept-licenses='android-sdk-license-[0-9a-f]{8}'
```
jverbraeken: If we use 19 as minimum sdk version, we should test with android 19 I think. Higher SDK versions generally do not remove functionality, but only add them

.travis.yml
ericcornelissen: It means only the last 5 commits are cloned

PR #291, opened 2017-06-18
IBAN for contacts

clenrock:
- Relevant Issues: - #241 
- Related Pull Requests: -

* * *

## What

This Pull Request adds to the repository a new field in Contact, and the ways to see them and change them in the ContactActivity.

## Why

This Pull Request is needed because the IBAN number is needed for integration.

## How

This feature can be viewed/tested within the project by going to the ContactActivity and the ChangeContactActivity.

## Alternative implementation

Other implementations that I've have considered are ...

## Notes

None

Issue #241, opened 2017-06-09
Integrate the app with the banking group


jverbraeken:
- Parent issues: -
- Relevant issues: -
- Estimated effort: 40

* * *

## User story
As a developer I want to check if we can integrate another application, namely the one made by the banking group under the lead of dhr. Pouwelse, in NervousFish so that we can show the functionality of NervousFish in a real-life scenario.

## What
This issue adds to the repository a check and, if it's possible, integration of the app of the banking group in NervousFish.

## Why
This issue is needed for the project because it would be very cool if we can demonstrate the functionality of NervousFish by integrating a third-party application.

## How
This feature can be added by talking with the banking group and evaluating if it's possible. If so, we'll have to think about how we want to do that.

## Definition of ready
This features is not ready until the following issues are finished:

- 

## Definition of done
This task can be considered done when

- We know if it's possible without too much trouble to integrate the application developed by the banking group in Nervous Fish
- If so, the application developed by the banking group is integrated in Nervous Fish

PR #292, opened 2017-06-19
[MERGE #291 FIRST] IBAN for Profile 

clenrock:
- Relevant Issues: -#241
- Related Pull Requests: -#291

* * *

## What

This Pull Request adds to the repository a new field in Profile and ways to add to that field.

## Why

This Pull Request is needed because you should be able to set your own IBAN.

## How

This feature can be viewed/tested within the project by removing and installing the app, and creating a new account. Here you can add an IBAN number. Also in the settings screen this can be edited.

## Notes

None


clenrock:
Closed because we don't use a string for IBAN anymore, but the IBAN object from the Bankver library

PR #245, opened 2017-06-12
Fix errors in QR and add some nice popups

clenrock:
- Relevant Issues: 
- Related Pull Requests: #224 

* * *

## What
This Pull Request adds to the repository some error fixes and new user friendly popups for QR exchange

## Why
This Pull Request is needed because there were sometimes ugly errors in QR.

## How
This feature can be viewed/tested within the project by exchanging a key via QR.

PR #297, opened 2017-06-19
Improve use of ContactReceivedHelper

ericcornelissen:
- Relevant Issues: -
- Related Pull Requests: -

* * *

## What

This Pull Request updates the repository so the `ContactReceivedHelper` is used mainly in the `MainActivity` and not throughout the Bluetooth activity and separately in the `NFCActivity`. Besides that it also adds a popup for when the contact is new and added to the database.

Also updated the documentation of the `ContactReceivedHelper`.

## Why

This Pull Request is needed because it improves consistency throughout the app.

## How

This feature can be viewed/tested within the project by exchanging contacts through Bluetooth of NFC and seeing that the popups work correctly.
~~Be careful with QR, even though the code has been changed, QR itself doesn't work correctly (it doesn't actually exchange contacts atm), wait or #245 before this works.~~ QR should also work, but it doesn't lead back to the `MainActivity`, this is because you want to allow the other to scan your QR as well (the popup shows up in the QRKeyExchangeActivity`).

## Alternative implementation

None considered

## Notes

None


app/src/main/java/com/nervousfish/nervousfish/activities/ContactReceivedHelper.java
```diff
@@ -41,17 +45,22 @@
      * @param contact  The newly received {@link Contact}
      */
     static void newContactReceived(final IDatabase database, final Activity activity, final Contact contact) {
+        if (contact == null) {
```
jverbraeken: Validate.notNull(database);
Validate.notNull(activity);
Validate.notNull(contact);
Same amount of lines, but 3 times more checks

app/src/main/java/com/nervousfish/nervousfish/activities/ContactReceivedHelper.java
```diff
@@ -41,17 +45,22 @@
...
+            return;
+        }
+
         switch (checkExists(database, contact)) {
             case NOT:
                 LOGGER.info("Adding contact to database...");
                 try {
                     database.addContact(contact);
+                    handleNewContact(activity);
```
jverbraeken: this

app/src/main/java/com/nervousfish/nervousfish/activities/ContactReceivedHelper.java
ericcornelissen: OK, but if you actually read the code you'd have noticed that `contact` is allowed to be `null` 😉

app/src/main/java/com/nervousfish/nervousfish/activities/ContactReceivedHelper.java
ericcornelissen: Was just following the style used in this class, but I'll change it 🙂

app/src/main/java/com/nervousfish/nervousfish/activities/ContactReceivedHelper.java
jverbraeken: I do not agree with that. The only way that contact can be null, is when your partner send you instead of his own Profile a null. That should not be possible, otherwise the time-consuming public key exchange is pretty useless.

ericcornelissen: Well, since all communication is within our application that should never happen 🤔 The reason I did it was so that you don't have to do this check manually [in the MainActivity](https://github.com/ericcornelissen/NervousFish/pull/297/files#diff-149538cc1f1f5e4370b6ae93d6489330R108), and potentially other activities. But if you prefer to have it changed, I'm OK with that 😄

PR #306, opened 2017-06-20
Fix duplicate contact with same keys received

jverbraeken:
- Relevant Issues: -
- Related Pull Requests: -

* * *

## What

This Pull Request adds to the repository a fix for when the user enters after receiving a contact with a duplicate name the name of another contact that also exists and thereafter types the name of another contact that's exactly the same as the new contact received. In that case nothing has to be done.

## Why

This Pull Request is needed because the app didn't recognize that the two contacts were the same

## How

This feature can be viewed/tested within the project by ...

## Alternative implementation

Other implementations that I've have considered are ...

## Notes

Needs testing!


clenrock:
The current way is just fine, and this does not make it more user friendly. (Discussed with Joost)

Issue #255, opened 2017-06-14
Add conditional tests for NFC/Bluetooth button on devices that support it


ericcornelissen:
- Parent issues: -
- Relevant issues: -
- Estimated effort: 2 hours

* * *

## User story
As a developer I want some way of automatically verifying that the Bluetooth/NFC fab buttons work even if it is not covered by any CI.

## What
This issue demands that tests are added to the `MainActivitySteps` that only run when Bluetooth/NFC are available on the device on which the tests are ran. Otherwise the tests should be skipped/do nothing.

## Why
This issue is needed for the project because it provides the minimum possible guarentee that these buttons work and do not crash the app. Some people (including myself) tend to run androidTests on their mobile device (which is likely to have Bluetooth/NFC).

## How
This feature can be added by creating new Scenarios in the `MainActivity.feature` file where the step definitions only conditionally (if Bluetooth/NFC are available) verify behavior and otherwise do nothing/don't fail.

## Definition of ready
n/a

## Definition of done
This task can be considered done when
- [x] When there are test(s) that verify the behavior of the Bluetooth fab button.
- [x] When there are test(s) that verify the behavior of the NFC fab button.
- [x] When all of the tests listed above are ignored if their respective feature is not available on the device where the tests are ran.

## Notes
None

ericcornelissen:
A test for the NFC button has been created with PR #311 🎉 

However, Bluetooth proofs rather difficult 😕 This is because, when you enter the `BluetoothConnectionActivity`, you are prompted to set your device status to discoverable, however I've not been able to control this popup through the automated androidTests...

ericcornelissen:
With commit c2305b9 on #311, there now also is a test for the Bluetooth button 🎉 It is less restrictive then the NFC test, since it only tests whether or not the app crashes. By doing that the "you are prompted to set your device status to discoverable"-problem has been circumvented, providing a minimum level of certainty the button is working.

PR #311, opened 2017-06-20
Add conditional test for the NFC & Bluetooth button in the MainActivity

ericcornelissen:
- Relevant Issues: #255 
- Related Pull Requests: -

* * *

## What

This Pull Request adds to the repository a conditional androidTests that verifies if the NFC button in the `MainActivity` is working as expected. It works as follows:
- If NFC is not available (i.e. the button is disabled) it does nothing.
- If NFC is available but not enabled, the test will click the button and select no in the popup, verifying that the activity doesn't change.
- If NFC is available and enabled, the test will go to the NFC activity and verify this.

## Why

This Pull Request is needed because it provides some certainty that the NFC button is actually working (assuming at least one person with an NFC enabled phone runs the android tests every now and then).

## How

This feature can be viewed/tested within the project by running the androidTest in an emulator and on your mobile device and see if the tests behave as expected.

## Alternative implementation

Other implementations that I've have considered are related to creating tests where the state of NFC is manually configured, however this proved impossible (?, at least very hard) to do because afaik you can't control other apps/the android settings through androidTests.

## Notes

None


app/src/androidTest/java/com/nervousfish/nervousfish/test/MainSteps.java
```diff
@@ -69,22 +71,17 @@ public void createDatabase() throws Exception {
...
         this.mActivityRule.launchActivity(intent);
-        allowPermissionsIfNeeded();
       
         try {
             onView(withText(R.string.no)).perform(click());
-        } catch (NoMatchingViewException ignore) {
-        }
-
-        allowPermissionsIfNeeded();
+        } catch (NoMatchingViewException ignore) { /* If no popup is displayed, that is OK */ }
```
TheBonheurs: Why is that ok?

app/src/androidTest/java/com/nervousfish/nervousfish/test/MainSteps.java
```diff
@@ -153,6 +170,28 @@ public void iShouldGoToSettingsScreen() {
         intended(hasComponent(SettingsActivity.class.getName()));
     }
 
+    @Then("^the app shouldn't crash because of Bluetooth$")
+    public void theAppShouldCrashBecauseOfBluetooth() {
+        assertTrue(true);
```
TheBonheurs: No way to test this I guess

app/src/androidTest/java/com/nervousfish/nervousfish/test/MainSteps.java
ericcornelissen: We don't know if Bluetooth is enabled (or even available) when the tests are run, so we don't know if a popup shows up requesting to enable Bluetooth. If it is there we don't want to enable it (because the permission might not yet have been given), and otherwise its OK 👍 We don't know if Bluetooth is enabled (or even available) when the tests are run, so we don't know if a popup shows up requesting to enable Bluetooth. If it is there we don't want to enable it (because the permission might not yet have been given), and otherwise its OK 👍

app/src/androidTest/java/com/nervousfish/nervousfish/test/MainSteps.java
ericcornelissen: No, the problem is the Android popup requesting the user to let the device be searchable for X amounts of seconds. Espresso is (afaik) not able to click on that popup and as long as its there we can't verify anything. Resulting in the test case error due to a timeout 😕

Issue #278, opened 2017-06-16
Prevent activities from going back to themselves


ericcornelissen:
- Parent issues: #234
- Relevant issues: -
- Estimated effort: 2 hours

* * *

## User story
As a user I want to go directly to the `MainActivity`/`BluetoothConnectionActivity` when I cancel a Bluetooth exchange process (at any point in the process)

## What
This issue demands that canceling/pressing back while in a exchange process with another phone will cancel the connection and lead back to the `MainActivity`/`BluetoothConnectionActivity`

## Why
This issue is needed for the project because it improves user experience and makes clear the connection has been canceled.

## How
This feature can be added by (proposed solution, not sure if this works) adding `noHistory` to the activities involved in the Bluetooth exchange process.

## Definition of ready
Either done when #249 is merged, or this issue can be combined with #249

## Definition of done
This task can be considered done when

- [ ] It is no longer the case that you go back to a previous activity when in the Bluetooth exchange process.
- [ ] It is still possible to start a connection directly after you just canceled a connection, even with the same device.

## Notes
None

PR #318, opened 2017-06-22
Add Bluetooth cancel functionality

ericcornelissen:
- Relevant Issues: #278
- Related Pull Requests: -

* * *

## What

This Pull Request adds to the repository functionality to actually cancel the Bluetooth connection. Either by pressing the Cancel button in the `WaitingActivity` or the back button in any of the activities that appear in the Bluetooth connection process

## Why

This Pull Request is needed because it makes it possible to cancel a Bluetooth connection whenever a user wants to. Besides that it also makes it so the actual connection is stopped, not just the activities.

## How

This feature can be viewed/tested within the project by starting a Bluetooth connection and trying to cancel it from every activity. Be sure to check if you can immediately start a new connection, as this shows that the previous connection has been canceled.

## Alternative implementation

None considered

## Notes

Not sure if `this.serviceLocator.getBluetoothHandler().stop();` is the correct way to stop a Bluetooth connection 🤔


jverbraeken (approved):

jverbraeken:
@ericcornelissen Did you test if this code works?

ericcornelissen:
I did not 😅 Sorry, should've been more clear about that

TheBonheurs (approved):

clenrock (changes requested):
Does not work as intended yet

PR #327, opened 2017-06-23
Delete existing contact popup

clenrock:
- Relevant Issues: -
- Related Pull Requests: -

* * *

## What

This Pull Request deletes the popup that shows when a contact already exists but the key is new.

## Why

This Pull Request is needed because it didn't work.

## How

This feature can be viewed/tested within the project by looking at the code.

## Notes

We should probably and fix this code after the deadline.


jverbraeken:
@clenrock Did you test this?

clenrock:
@jverbraeken I love the question! Yes, it is tested and works very well

PR #331, opened 2017-06-25
Get Ed25519 working with the BankBoys dependency and serializable

ericcornelissen:
- Relevant Issues: #241 
- Related Pull Requests: -

* * *

## What

This Pull Request adds to the repository an update for the Ed25519Key that allows it to be used in the BankBoys dependency while still keeping it serializable.

## Why

This Pull Request is needed because we want both of these things at the same time.

## How

This feature can be viewed/tested within the project by using the Ed25519 key in some way and see if the app still works

## Alternative implementation

None considered

## Notes

None


clenrock (approved):

TheBonheurs (approved):
Great work!

PR #339, opened 2017-06-26
Show an explanation in the rhythm / visual verification screen

jverbraeken:
- Relevant Issues: -
- Related Pull Requests: -

* * *

## What

This Pull Request adds to the repository an explanation in the RhythmCreateActivity and VisualVerificationActivity

## Why

This Pull Request is needed because it was not clear at all for a test audience what they had to do in the rhythm and visual verification activity.

## How

This feature can be viewed/tested within the project by ...

## Alternative implementation

Other implementations that I've have considered are ...

## Notes

The available space to tap is now a little bit smaller in visual verification activity. If this is really a problem, then we should move the image below the explanation text.

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -106,6 +106,7 @@
     <string name="not_the_same_rhythm_explanation">The rhythm you tapped was not the same as on the other device.</string>
     <string name="not_the_same_visual_title">Tapping order was the same</string>
     <string name="not_the_same_visual_explanation">Make sure that the locations tapped and the order are the same on both devices.</string>
+    <string name="rhythm_explanation">Press \"Start recording\" and tap the same rhythm as your partner</string>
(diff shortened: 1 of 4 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -41,6 +41,7 @@
     <string name="something_went_wrong">Something went wrong</string>
     <string name="try_again">Please try again.</string>
     <string name="waiting_for_partner">Waiting for partner…</string>
+    <string name="visual_verification_explanation">Tap 6 times in the same order on objects as your partner</string>
```
clenrock, line 44: Dutch description is missing

jverbraeken: Fixed

TheBonheurs (approved):

clenrock (approved):

Issue #315, opened 2017-06-21
Use the custom keyboard throughout the application


ericcornelissen:
- Parent issues: -
- Relevant issues: #29
- Estimated effort: 5 hours

* * *

## User story

As a user I want to have a custom keyboard in the NervousFish app so that I can be sure noone is listening to what I'm doing

## What

This issue demands that the custom keyboard (which is already implemented, and is being used in the `LoginActivity) is used throughout the app for all inputs.

## Why

This issue is needed for the project because we want user to be sure it is secure to use our app, and not even their keyboard is spying on them.

## How

This feature can be added by _Unclear as of now_

## Definition of ready

This features is not ready until the following issues are finished:
- Waiting on other issues is not needed

## Definition of done

This task can be considered done when:
- [ ] The custom keyboard is used in the settings
- [ ] The custom keyboard is used in the edit contact activity
- [ ] The custom keyboard is used in the create profile activity

## Notes

If I missed any places where the custom keyboard should be used, be sure to comment and I'll add it to the list :smile:

PR #346, opened 2017-06-30
Implement shift key for custom keyboard

justinvdk:
- Relevant Issues: #315 
- Related Pull Requests: n/a

* * *

## What

This Pull Request adds to the repository a shift key for the custom keyboard.

## Why

This Pull Request is needed because without shift key users cannot use capital letters in their password for now, and with closing #315 nowhere.

## How

This feature can be viewed/tested within the project by opening custom keyboard (loginActivity as of now) .

## Alternative implementation

Other implementations that I've have considered are None. Seems like the correct solution.

## Notes

The unicode character used (u21e7) should be correct, but seems out of order. This is up to maintainers to accept or not.


justinvdk:
Changed base branch

TheBonheurs (approved):
Very bad coding style

ericcornelissen (approved):

